PR #40, opened 2013-11-29
Interactive Execute command 

miry:
Hi @leehambley 

To add ability to run interactive rails console before for capistrano 2 we could do:

``` ruby
desc "Remote console on the production appserver"
task :console, :roles => :app do
  input = ''
  run "cd #{current_path} && ./script/console production" do |channel, stream, data|
    next if data.chomp == input.chomp || data.chomp == ''
    print data
    channel.send_data(input = $stdin.gets) if data =~ /^(>|\?)>/
  end
end
```

this solution does not work for v3, and I could not find similar.
I have patched a SSHKit to execute interactive rails console:

https://gist.github.com/miry/7196069.

Can you advice me another solution?

Also I think to store the channel and then we can redefine `on_data` or another callback method, or add block to `execute` method with same functionality like old `run`.

### DIFF
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -54,16 +54,16 @@ module SSHKit
         instance_exec(host, &@block)
       end
 
-      def test(*args)
+      def test(*args, &block)
         options = args.extract_options!.merge(
           raise_on_non_zero_exit: false,
           verbosity: Logger::DEBUG
         )
-        _execute(*[*args, options]).success?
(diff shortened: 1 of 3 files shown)


afeld:
I blatantly stole @miry's work and made a standalone gem, [SSHKit::Interactive](https://github.com/afeld/sshkit-interactive). Would love [feedback](https://github.com/afeld/sshkit-interactive/issues/3).

miry:
@afeld nice :+1:

PR #88, opened 2014-01-03
Add Netssh#stream to give real-time access to stdout & stderr

lmars:
The main reason for this is so that you can print STDOUT as it is received.

A contrived example:

``` ruby
on "root@example.com" do
  cmd = 'for i in $(seq 20); do echo "the time is $(date +%H:%M:%S)"; sleep 1; done'

  stream(cmd) do |type, data|
    print data
  end
end
```

### DIFF
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -76,6 +76,11 @@ module SSHKit
         _execute(*[*args, options]).full_stdout.strip
       end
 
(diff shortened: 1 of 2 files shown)


JeanMertz:
@leehambley since this wasn't merged, how can one do `tail -f` streaming? Whenever I try `execute` with `tail -f` I don't get any streamed data.

leehambley:
Use @lmars' fork if you like.

PR #134, opened 2014-04-27
Add gateway support.

msimonin:
When a gateway is declared (globally or per host), a connection is created to this gateway to encapsulate
all the connections to the hosts using this gateway.

Add hash SSHKit::Backend::ConnectioPool.connections instead of using Thread.current hash.
This allows for multiple threads requesting connections to the same host to reuse the same connections.

Add SSHKit::Host.gateway

Add SSHKit::Backend::Netssh::Configuration.gateway

M

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -363,3 +363,52 @@ Replace `on` with `run_locally`
       end
     end
 
+## Using a gateway
+
+SSHKit allow you to use gateway(s). 
+All subsequent connections will be tunneled through the gateway(s) (using SSH forwarded ports). To tell SSHKit about your gateways you can either :
+
+* define a global gateway
+
+```ruby
+SSHKit.config.backend.configure do |ssh|
+    ssh.gateway = "user@my-gateway"
+end
+```
+
+The gateway definition follows the same syntax as host definition. This means you can define your global gateway like this:
+
+```ruby
+SSHKit.config.backend.configure do |ssh|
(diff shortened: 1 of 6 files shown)


leehambley:
@mattbrictson since you bumped this thread from the other side, I defer to your knowledge a bit of whether we can/should do this? I believe the problem is resolved with project-local ssh configuration, but I also don't use/need gateways.

mattbrictson:
@leehambley Yes, I agree that project-local SSH configuration can solve this. I regularly deploy to a production environment that requires a gateway, and it works fine with an SSH configuration that looks like this:

```
Host gateway
  ForwardAgent yes
  HostName <GATEWAY_IP>

Host prod-app01
  ProxyCommand ssh gateway exec nc <APP_PRIVATE_IP> %p

Host prod-db01
  ProxyCommand ssh gateway exec nc <DB_PRIVATE_IP> %p
```

leehambley:
Thanks @matt, I was sure it worked. I'm not sure how Windows works, but given that we pass this file to Net::SSH anyway, I can assume (anyone can confirm?) that it does the right thing.

dalehamel:
@mattbrictson under the current implementation, and the one you propose, it seems that capistrano leaks hundreds of processes and file descriptors if you have more than a dozen or so hosts behind your gateway. 

It would be ideal to be able to reuse the gateway rather than waste these resources, i don't think this PR should be closed as it seems to do exactly that.

PR #156, opened 2014-07-08
Add quotes around environment variable values when generating environment_hash

keithpitt:
With something like this:

```ruby
with upload_url: "http://foo.com?bar=foo&bar=foo" do
  ...
end
```

The `environment_string` generates a string that looks like this: `UPLOAD_URL=http://foo.com?bar=foo&bar=foo` which is invalid because it contains the ampersand (`&`). The solution I found was to wrap the values in quotes.

Was this an acceptable solution?

### DIFF
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -152,7 +152,7 @@ module SSHKit
 
     def environment_string
       environment_hash.collect do |key,value|
-        "#{key.to_s.upcase}=#{value}"
+        "#{key.to_s.upcase}=\"#{value.to_s.gsub(/"/, "\\\"")}\""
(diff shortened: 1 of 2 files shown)


leehambley:
Is this fixed by https://github.com/capistrano/sshkit/pull/250? Can you check against the master branch please?

leehambley:
No feedback from @keithpitt for 10 months

keithpitt:
Ah sorry! I wasn't subscribed to the issue and I didn't get an @keithpitt notification. I didn't actually know this was fixed. Thanks!

leehambley:
No worries @keithpitt thanks for coming back to me!

PR #196, opened 2014-12-15
Fixed documentation

Stas-Ghost:
Parameter -d  return true value if path exists and is a directory. So for file we should use -f.
[ -f /some/file ] returns true when file exists, so statement branches should be switched.


mattbrictson:
@Stas-Ghost if you are still around, could you fix the conflicts in this PR so I can merge it? Thanks.

Stas-Ghost:
Conflicts fixed

PR #211, opened 2015-01-07
Fix pretty logger log level aligning

mikz:
* this reverts #185 as it did not work
* adds a test and justifies all log levels on one place


sponomarev:
:+1:

PR #212, opened 2015-01-07
introduce SSHKit.config.local_backend configuration

mikz:
* ability to switch local backend
* allows to perform dry run on local

This would allow capistrano to do following:

```ruby
SSHKit.configure do |config|
  config.backend = Capistrano.env.get(:sshkit_backend)
  config.local_backend = Capistrano.env.get(:sshkit_local_backend)
end
```

So when doing dry run, capistrano would set local backend to `LocalPrinter` instead of `Local`.
Because now same `Local` backend is used for dry run as normal run. That results in weird
effects like running deploy notifications when doing dry-run.

### DIFF
--- a/lib/sshkit/all.rb
+++ b/lib/sshkit/all.rb
@@ -30,6 +30,7 @@ require_relative 'backends/abstract'
 require_relative 'backends/connection_pool'
 require_relative 'backends/printer'
 require_relative 'backends/netssh'
+require_relative 'backends/local_printer'
 require_relative 'backends/local'
 require_relative 'backends/skipper'
 
--- a/lib/sshkit/backends/local.rb
+++ b/lib/sshkit/backends/local.rb
@@ -1,19 +1,12 @@
 require 'open3'
 require 'fileutils'
+require 'sshkit/backends/local_printer'
+
 module SSHKit
 
   module Backend
 
-    class Local < Printer
-
-      def initialize(&block)
-        @host = Host.new(:local) # just for logging
-        @block = block
-      end
-
-      def run
-        instance_exec(@host, &@block)
-      end
+    class Local < LocalPrinter
 
       def test(*args)
         options = args.extract_options!.merge(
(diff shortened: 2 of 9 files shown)


niku4i:
:+1: I was looking for this feature.

Any updates on this? It looks conflicts though.

mikz:
@niku4i I can fix the conflicts if it gets :+1: from the maintainers

leehambley:
I don't recall having seen this PR before, this is explicitly about a "Local PRinter" backend, that is specifically "local" but…"printing" … I don't see why it's not just "Printer"? Probably I missed something.

Short historical note, the `dry run` was removed from Cap because for all but the trivial cases, there's always *something* like, a `capture()` or something date/time related, where the old "printer" backends would just fail to proceed.

mikz:
@leehambley because the printer does not evaluate the block, in this case local printer does.

So if dry run was removed, how you test deployment?

leehambley:
> So if dry run was removed, how you test deployment?

Vagrant environment or staging, it has been our experience that anything else falls short of representing the real state of affairs.

My own teams are lucky that almost exclusively we can share CM configs between staging and production, *and* with our development environments. 

I understand this isn't always true, but I'd rather then recommend to deploy your staging environment and use something like http://serverspec.org/ to test that things look as expected.

sonots:
:+1: want dry-run

leehambley:
In other words I could accept this PR, but I'd like to understand better (can we solve it by naming) then the difference between all the "Backends" and if you've already done some work on that, maybe we can add as part of this PR a table to the README explaining the use-cases for the ones we ship.

leehambley:
Also, it occurs to me that thanks to the Phusion guys we can do `on(:local)` which uses the local runner, so then can't the regular printer backend be used?

mikz:
There were some significant changes that possibly make this PR invalid. Will try to take a look, rebase, ...
And checking it after some time, the naming sucks. I don't get what it is supposed to do :)

leehambley:
> I don't get what it is supposed to do :)

I think it's supposed to be a precursor to making `--dry-run` work with the local backend, which currently always runs.

mattbrictson:
I'm in the process of closing old/inactive PRs; pardon the noise. This PR seems to have been abandoned. Please comment if you'd like it to be reconsidered.

Issue #203, opened 2014-12-19
SSHKit::Runner::ExecuteError: Exception while executing as deploy@target.ch: no implicit conversion of Hash into String


raskhadafi:
After `bundle update` which updated all outdated gems. See the list.

* capistrano (3.3.4 -> 3.3.5)
* colorize (0.7.4 -> 0.7.5)
* sshkit (1.5.1 -> 1.6.1)

Following error occured.

```bash
** Invoke application:staging (first_time)
** Invoke capwodi:ensure_latest_version (first_time)
** Execute capwodi:ensure_latest_version
** Execute application:staging
** Invoke load:defaults (first_time)
** Execute load:defaults
** Invoke bundler:map_bins (first_time)
** Execute bundler:map_bins
** Invoke check:default (first_time)
** Invoke check:app_user (first_time)
** Execute check:app_user
** Execute check:default
** Invoke rvm:hook (first_time)
** Execute rvm:hook
cap aborted!
SSHKit::Runner::ExecuteError: Exception while executing as deployer@server.ch: no implicit conversion of Hash into String
/User/home/me/gems/colored-1.2/lib/colored.rb:70:in `colorize'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/color.rb:7:in `black'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/formatters/pretty.rb:67:in `level'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/formatters/pretty.rb:22:in `write_command'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/formatters/pretty.rb:10:in `write'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/backends/netssh.rb:134:in `block in _execute'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/backends/netssh.rb:133:in `tap'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/backends/netssh.rb:133:in `_execute'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/backends/netssh.rb:62:in `test'
lib/tasks/cap/rvm.cap:19:in `block (3 levels) in <top (required)>'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/backends/netssh.rb:54:in `instance_exec'
/User/home/me/gems/sshkit-1.6.1/lib/sshkit/backends/netssh.rb:54:in `run'
[...]

PR #213, opened 2015-01-07
Require colorize version < 0.7.5

chopmo:
As described in #203, upgrading to `colorize` 0.7.5 breaks something in sshkit. 

I'll admit that I haven't looked into the root cause, but I thought this might be an acceptable solution for now.

### DIFF
--- a/sshkit.gemspec
+++ b/sshkit.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |gem|
 
   gem.add_runtime_dependency('net-ssh',  '>= 2.8.0')
   gem.add_runtime_dependency('net-scp',  '>= 1.1.2')
-  gem.add_runtime_dependency('colorize', '>= 0.7.0')
+  gem.add_runtime_dependency('colorize', ['>= 0.7.0', '< 0.7.5'])
 
   gem.add_development_dependency('minitest', ['>= 2.11.3', '< 2.12.0'])
   gem.add_development_dependency('rake')


leehambley:
Thanks, but there's a better solution, which I thought was already merged, I will check and report back (and possibly close) here when I know more.

leehambley:
This commit ought to have fixed that 19052837cf91b044d80e466d1a6dfeb826b40462, which is included in the 1.6.1 release https://github.com/capistrano/sshkit/tree/v1.6.1

chopmo:
So sorry for the late reply @leehambley, completely missed this one. 

I'm still seeing the issue with sshkit 1.7.1 and colorize 0.7.5. 

[Here](https://gist.github.com/chopmo/df79858c68841b13e1eb) is the backtrace. 

Let med know if I can do more to help and I'll make sure to repond quickly.

mattbrictson:
@chopmo Based on your backtrace, the error is coming from [colored](https://github.com/defunkt/colored), not colorize. It seems you have both gems loaded, and each is trying to monkey patch String in different ways. Can you find how/why colored is being required in your case?

```
TypeError: no implicit conversion of Hash into String
/opt/boxen/rbenv/versions/2.2.0/lib/ruby/gems/2.2.0/gems/colored-1.2/lib/colored.rb:70:in `colorize'
                                                         ^^^^^^^^^^^
```

chopmo:
@mattbrictson You're right, good catch. 

`colored` was being required by [this cap-ec2 fork](https://github.com/AmirKremer/cap-ec2/tree/v0.0.17). I was using a fork of cap-ec2 because of a bug which is now fixed. 

So all in all, this issue can now be closed.

PR #215, opened 2015-01-22
Use popen3 for local backend

deees:
Fixes issue when long running local commands output to console at the end only. Local backend now "streams" stdout/stderr immediately to console.

@leehambley


deees:
thanks!

Issue #216, opened 2015-01-24
Control STDIN for individual command executions


vivid-inc:
Feature request! Two humble, rudimentary use cases, the abstractions of which are of high value to us:

1) A laundry list of system admin commands is written in a Ruby script, almost all of which require privileged access to run properly. A sudoers.d configuration file permits these certain system maintenance commands to be executed without password; a dangerous practice. Instead, with SSHKit, scripts can start with `sudo -v' to update the user's cached credentials, and subsequent commands have a plain `sudo' command-map prefix so that they can still run correctly. The key change being when `sudo -v' is run, the script can prompt the user for their password, subsequently piping that input to sudo's STDIN, something like this:

```ruby
execute(:sudo, '-v') do |context|
    password = STDIN.noecho(&:gets).chomp
    context.stdin.send_data(password) # Textual. `sudo' eventually next receives EOF, somehow.
end
```

2) Customer data is protected from intrusions by virtue of the data-at-rest being encrypted. That data is periodically processed, but must be decrypted first, and the corresponding secret key material must never be able to touch disk (assume swap is encrypted) or be visible (i.e. arguments in the process table). With SSHKit, the automation script can obtain the secret key from the local system and pipe it into the program's STDIN, thereby reducing the potential for ... mishaps on the remote system.

```ruby
execute(:process_customer_data, "--secret_key=- args ...") do |context|
    context.stdin.send_data(secret_key) # Binary blob. Again, STDIN EOF somehow follows
end
```

Thanks for your hard work on the fabulous SSHKit. We are looking at rewriting all automation scripts with something, and this feature would seal the deal for SSHKit.

leehambley:
I'd be more than happy to merge this if you can also add something to EXAMPLES and to the readme. I'd appreciate if it was worded to apply caution.

I'd also feel *much* better if you could add some tests, using `StringIO` to test that in a simple way would be ideal, even if you just `cat somefile` and write tostdin, and assert that the file has the correct contents.

The test suite spins up real VMs.

robd:
Hi @vivid-inc 

I'm looking at adding support for user interactions in #234. I'm pretty sure it will support the sudo use case you mention and I think it could support the `process_customer_data` case too. Would you be happy to close this issue if #234 was merged?

Any feedback much appreciated.

vivid-inc:
[...]

PR #217, opened 2015-01-25
Controlling STDIN

cfis:
Allow calling methods that use execute to process the data returned by an ssh command and respond on stdin if needed.  One approach to addressing #216.


leehambley:
People can use Net::SSH if they want to have access to the raw connection. I think this confuses the API, although I do concede that access to the underlying file descriptors would be nice from time to time.

PR #219, opened 2015-01-26
Do not include SSHKit::DSL in the context it is required in

beatrichartz:
In https://github.com/capistrano/sshkit/pull/28 the dsl file was removed from the original requires - I assume since the DSL was also automatically included in the context it was required in.

If I want to load the DSL _and_ use it in another context than required in, the current inclusion pattern for the DSL leaves no options - the DSL gets auto-included wherever I require it.

This leads to code like:

```ruby
require 'sskit'
module MyModule
  class ClassUsingDSL
    require 'sshkit/dsl'
    #...
  end
end
```

which is suboptimal. We can require the dsl with all other files and make the include optional instead, which allows for:

```ruby
require 'sshkit'
module MyModule
  class ClassUsingDSL
    include SSHKit::DSL
    #...
  end
end
```

The required changes are in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ This file is written in reverse chronological order, newer releases will
 appear at the top.
 
 ## `master` (Unreleased)
+  * Do not auto-include DSL in context of `require`. Load DSL with Gem and allow to include it.
+    [PR #219](https://github.com/capistrano/sshkit/pull/219)
+    @beatrichartz
   * `SSHKit::Backend::Netssh.pool.idle_timeout = 0` doesn't disable connection pooling anymore,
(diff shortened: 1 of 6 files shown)


leehambley:
@mattbrictson any reasons you see not to merge this? I missed it, but it seems like a win.

mattbrictson:
@leehambley This looks good to me. I was worried about this breaking Capistrano, but it looks like Capistrano never actually uses the SSHKit::DSL; instead it defines its own `on` method that does something similar. So I think that means we are safe.

We need a CHANGELOG and a rebase for this PR.

leehambley:
@beatrichartz could you take care of that please?

beatrichartz:
@leehambley done

PR #221, opened 2015-02-11
Images in 'lib/assets' are incorrectly picked up as Rails' sprocket assets.

phallstrom:
Having example images in 'lib/assets' makes it appear as those are "Rails like" assets for use in Sprockets.  Specifically Middleman app looks for this directory and includes it in the build output. As these images aren't assets, but simply used in the README they should be moved into an alternate directory to avoid confusion.

See here for what Middleman (and I would assume other sprocket aware gems) look for:

https://github.com/middleman/middleman-sprockets/blob/master/lib/middleman-sprockets/extension.rb#L119

There is ZERO FUNCTIONAL CHANGE to the gem itself.  Only the README is affected.


leehambley:
We don't build with either middleman or sprockets, is this really worth fixing? (Sorry for my ignorance, but we're rendered with Github's README/markdown parser, so I don't see the value of "fixing" this thing that isn't broken?)

phallstrom:
You don't build with those, but I am using capistrano to deploy a middleman app that uses sprockets.  Capistrano requires sshkit.  With this setup I pick up those images in my production build and can't prevent it.

It's broken in the sense that if any app uses sshkit and sprockets in the right combination it may pick those images up which isn't what should be happening.

Right?

leehambley:
I'd put `capistrano` in a `development` group of your Gemfile, or don't vendor it. I'll merge the PR because it's harmless, but if parts of Cap and SSH kit are showing up in your build/project dir, I think you might have wider problems! Good luck, and thanks for the contrib!

phallstrom:
Appreciate it.  Sadly capistrano is in my development group.  Just has to do with how middleman builds it and how you have to invoke it.

I imagine the thinking went "people will add the jquery, bootstrap, etc. etc." gems which are full of assets and it would be nice to automatically include them.  Heh.

ETA on a rubygem cut?

leehambley:
Fair enough, but adding "jquery" as a Gem is flat out plain stupid, however
you cut it :) But, nobody said we had to agree with the ecosystem, right :)

Lee Hambley
http://lee.hambley.name/
+49 (0) 170 298 5667

phallstrom:
Haha!  Very true.  There are others though.. compass maybe.  I dunno.  I don't do it either, but it happens :)

PR #224, opened 2015-02-19
Fix duplicate messages

townsen:
Reset command stdout and stderr strings to avoid duplicate messages
Introduced by the race condition fix

Sorry about this! I didn't spot that the DEBUG messages for stdout and stderr were being duplicated..


leehambley:
I should have checked, the duplication (as you probably now see) was my (lame) attempt at not losing it in case (as I didn't know then) that the signals arrive out of order. 

Note to self: Don't raise exceptions on assignment, that was a really stupid API decision, at least we can still fix that 
:)

PR #225, opened 2015-02-23
Place full output in exception on failed command

townsen:
Fix omission in an earlier fix for failed command output so that
when an error occurs all the stderr/stdout is placed into the exception


leehambley:
Thanks Nick

PR #227, opened 2015-03-02
Log command *after* setting its exit_status

### LINKED ISSUE #226: Regression: SSHKit 1.7.0 removes command completion status from logs
In SSHKit 1.6.0 the successful completion status of each command was logged, like this:

```
INFO [b02151f4] Finished in 0.141 seconds with exit status 0 (successful).
```
In SSHKit 1.7.0 these log messages are gone. SSHKit 1.7.0 logs the command but never logs the "Finished" message.

This regression is due to this commit: https://github.com/capistrano/sshkit/commit/aa278dbeb4b1f931c875d413a22b22971085189c#diff-6285ece7ab8d7c6b765599fa5ee02b01

Before, the `exit_status` was updated first, and then the command was logged, like this:

```ruby
cmd.exit_status = data.read_long
output << cmd
```

Now in 1.7.0, the order is flipped, such that the command is logged *before* the status is updated:

```ruby
output << cmd
cmd.exit_status = data.read_long
```

As a result, the logger never receives the command with the exit status. This is very important information that is now missing from the log output.

mattbrictson:
This restores SSHKit 1.6 behavior and ensures that command completion status messages appear in the log output.

Fixes #226.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ appear at the top.
 
 ## `master` (Unreleased)
 
+  * Fix a regression in 1.7.0 that caused command completion messages to be removed from log output. @mattbrictson
   * Add your entries here, remember to credit yourself however you want to be
     credited!
 
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -150,7 +150,6 @@ module SSHKit
                 end
                 chan.on_request("exit-status") do |ch, data|
                   exit_status = data.read_long
-                  output << cmd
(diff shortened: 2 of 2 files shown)


leehambley:
@mattbrictson any chance of a test for this?

mattbrictson:
I didn't see much precedent for testing Backend::Netssh. Do you have any pointers? I'd rather not create mocks for the entire net/ssh API that Backend::Netssh depends on. If you think there is an easier way I can certainly give it a shot.

leehambley:
Alright, I should have checked it.

PR #237, opened 2015-04-26
Simplify backend hierarchy

robd:
This PR also came out of my explorations for implementing #234. Specifically, I was looking at the possibility of supporting interactive sessions by subclassing, rather than adding another option to `Command`. I was looking to see where it was necessary to hook `Command.new` and as far as I can tell `execute` in [CommandHelper](https://github.com/capistrano/sshkit/blob/af6390432dc7da44afc166f8701ccd048323dec5/lib/sshkit/command.rb#L8) wasn't actually called because the the backends which included `CommandHelper` overrode `execute`. This led me to look into the backend hierarchy and try to simplify things a bit.

This PR includes 3 main changes:

1) Moved `make`, `rake`, `test`, `capture` and `background` on to the `Abstract` backend. This was to remove duplicate implementations on `Local` and `Netssh` backends and remove redundant inclusion of `CommandHelper` in Netssh.
2) Moved the following duplicate command creation code from the `Local`, `Netssh`, `Skipper` and `Printer` backends into the `Abstract` backend by implementing `execute` in `Abstract` and introducing the `execute_command` abstract method on the subclasses instead.

```ruby
def execute(args)
  command(*args).tap do |cmd|
    # Execute the command
  end
end
```

3) Removed `Printer` from the `Local` and `Netssh` hierarchies since the `Printer` method implementations were overridden by the subclasses. As part of this, I also moved the `output` method from `Printer` up to `Abstract`, since it is needed by the `info(messages)`, `debug(messages)` etc methods on `Abstract`. I also implemented `run` on `Abstract` and removed the duplicate implementations from the subclasses.


leehambley:
Thanks Rob :)

robd:
No probs - thanks for merging this so quickly. I hope to be able to actually implement #234 now!

PR #239, opened 2015-04-26
No longer strip whitespace/newlines in capture method on Netssh backend

robd:
This is a breaking change, which I wanted to discuss before merging - I'm not sure what the best course of action is here.

The old `capture` behaviour was to `strip` whitespace and newlines from the captured results. I think the main reason for this was to remove trailing newlines from command results to make one liners easier to handle (eg `capture(:uname)` returns `Linux` as opposed to `Linux\n`). Unfortunately this meant that you couldn't capture command output with trailing whitespace / newlines. This was discussed in #165 with @townsen, and I believe it was decided to make this change, however this was only changed for the `Local` backend in 7d15a9a, but not for the Netssh one. I think this was an omission, so, if possible, I would like to fix this inconsistency now.

From an API (purist?) perspective, I think it makes sense to remove the `strip` method from the `Netssh` backend too as per the original decision in #165 and this PR implements that. But this might cause quite widespread breakages in client code (although easily fixed as mentioned in #165). This might mean the need for a major release version bump - I'm not sure what the policy is here.

Another option would be to revert the change made in 7d15a9a (ie put back the `strip` behaviour on `Local` backend `capture`), and introduce a new method (eg `capture_raw`) which doesn't strip. Of course the downside of this is clutter on the `Abstract` backend API which must be maintained, but this would mean that people wouldn't have to change their client code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * No longer strip whitespace or newlines in `capture` method on Netssh backend. @robd
+    * This is to make the `Local` and `Netssh` backends consistent (they diverged at 7d15a9a)    
+    * If you need the old behaviour back, call `.strip` (or `.chomp`) on the captured string i.e. `capture(:my_command).strip`  
   * Simplified backend hierarchy. @robd
     * Moved duplicate implementations of `make`, `rake`, `test`, `capture`, `background` on to `Abstract` backend.
     * Backend implementations now only need to implement `execute_command`, `upload!` and `download!`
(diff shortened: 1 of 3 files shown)


leehambley:
Cheers @robd, I'll make an announcement via Gem post install message, or something when we do the release that this might introduce breaking changes, no need to bump the API though, I don't think.

townsen:
@robd, indeed I should have made this change for Netssh too, but it never occurred to me (I only had the problem locally with parsing 'git' output). I agree that 'capture' staying pure seems to be best, and as Lee has pointed, people should read the message and test before upgrading.

robd:
@townsen OK great - thanks for replying. I don't think anyone has actually come across the need for preserving whitespace with the Netssh backend (or at least not that I could find in the issues), so I don't think it was a problem, but good to have it consistent.

PR #242, opened 2015-05-04
Improve interaction handler api

robd:
This PR contains 4 improvements to the `:interaction_handler` option API from trying to use this with Capistrano in the real world:

1) Put the `MappingInteractionHandler` class in the `SSHKit` module
2) Support specifying an `interaction_handler` as a hash
```Ruby
# Before:
execute(:some_command, interaction_handler: SSHKit::MappingInteractionHandler.new("Some output\n" => "Input\n"))
# After:
execute(:some_command, interaction_handler: { "Some output\n" => "Input\n"})
```
3) No longer append a newline to the user input. This removes an inconsistency between the server side and user input side of the mapping:

```Ruby
# Before - newline needed on server output only because newline magically appended to input:
execute(:some_command, interaction_handler: { "Some output\n" => "Input"})
# After - no magic newline behaviour; all newlines should be specified:
execute(:some_command, interaction_handler: { "Some output\n" => "Input\n"})
```
4) Remove the`output` constructor parameter and attribute from `MappingInteractionHandler`. This was only added for unit testing purposes, but meant that it didn't work with [airbrussh](https://github.com/mattbrictson/airbrussh/commit/ec3122b101de53f2304723da842d5c8b6f70f4f3#diff-faf9277ec199ea9f2e313151176a085eR230). It also meant that `MappingInteractionHandler` would hold on to a reference to `SSHKit.config.output` which seems a bit dodgy to me in the case that `SSHKit.config.output` is updated after the `MappingInteractionHandler` is created.

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -80,12 +80,12 @@ have a shell defined, one cannot switch to that user.
 
 ```ruby
 on hosts do |host|
-  execute(:passwd, interaction_handler: MappingInteractionHandler.new(
+  execute(:passwd, interaction_handler: {
     '(current) UNIX password: ' => 'old_pw',
     'Enter new UNIX password: ' => 'new_pw',
     'Retype new UNIX password: ' => 'new_pw',
     'passwd: password updated successfully' => nil # For stdout/stderr which can be ignored, map a nil input
-  ))
+  })
 end
 ```
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -81,9 +81,9 @@ have a shell defined, one cannot switch to that user.
 ```ruby
 on hosts do |host|
   execute(:passwd, interaction_handler: {
-    '(current) UNIX password: ' => 'old_pw',
-    'Enter new UNIX password: ' => 'new_pw',
-    'Retype new UNIX password: ' => 'new_pw',
+    '(current) UNIX password: ' => "old_pw\n",
+    'Enter new UNIX password: ' => "new_pw\n",
+    'Retype new UNIX password: ' => "new_pw\n",
     'passwd: password updated successfully' => nil # For stdout/stderr which can be ignored, map a nil input
   })
 end
--- a/README.md
+++ b/README.md
@@ -243,19 +243,19 @@ execute(:passwd, interaction_handler: PasswdInteractionHandler.new)
 ```
 
 Often, you want to map directly from an output string returned by the server to the corresponding input string (as in the case above).
-For this case you can use a `SSHKit::MappingInteractionHandler`:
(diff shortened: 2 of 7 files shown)


leehambley:
Would a regex work here?

    execute(:some_command, interaction_handler: { /Some output\n+/ => "Input\n"})

robd:
@leehambley I have thought about adding support for this. At the moment, I simply treat the hash keys as hash keys so it won't work (I don't think, because hash uses eql? for key equality), but obviously I could add support for this. 

Another option would be to support a lambda:

```ruby
execute(:some_command, interaction_handler: lambda { |output_line|
  case output_line
  when /Some output\n+/
    "Input\n"
  end
)
```

I definitely think that some way to match a regex would be helpful. This `{ /Some output\n+/ => "Input\n"}` feels a bit non idiomatic, but is definitely much nicer for one liners.

robd:
I also added a fix for a change of behaviour I accidentally introduced to do with mutating strings which broke airbrussh in a non obvious way and forced @mattbrictson to implement a [pretty nasty workaround](https://github.com/mattbrictson/airbrussh/commit/ec3122b101de53f2304723da842d5c8b6f70f4f3#diff-faf9277ec199ea9f2e313151176a085eR230).

leehambley:
Oopse sorry @mattbrictson :) I missed the link with airbrussh when reviewing this :(

robd:
@leehambley I'm adding regex support now. Will update this PR soon.

robd:
OK, I now compare the hash keys using case equality to support regexes. I also support specifying a proc. I also changed the behaviour for unmapped commands to output at the debug level instead of warning level. 
```ruby
execute(:some_command, interaction_handler: { /Some output\n+/ => "Input\n"})

# OR

execute(:some_command, interaction_handler: lambda { |output_line|
  case output_line
  when /Some output\n+/
    "Input\n"
  end
)
```

One thing to note is that since the `on_stdout`/`on_stderr` are called once per stdout/stderr line, I think that `\n+` in your regex won't ever get a chance to match multiple newlines.

mattbrictson:
@robd Thanks! I tested this branch and it fixes the issue that required me to add the `deep_copy` workaround to airbrussh.

robd:
@mattbrictson OK great - thanks vm for checking

leehambley:
At first glance (can't review too thoroughly, or try your branch out) but this all looks sane, with such a flexible API making the docs clean and clear will be essential.

leehambley:
Happy to merge, if you are ?

robd:
Yes, I'm happy with this - thanks. I hope the Readme covers all usage patterns, but if anything crops up I will raise as a separate PR.

Issue #243, opened 2015-05-06
Support for assigning a stream to SSHKit.config.output is broken


robd:
Hey @leehambley 

I wanted to discuss this before I try to fix it, but I think this syntax, which is included in the Readme, is a bit broken. Do you know, is this supposed to work, or is the readme out of date?

```ruby
SSHKit.config.output = $stdout
```

I rigged up a demo script with sshkit 1.7.1 to see what the output actually looks like:

```ruby
require 'rubygems'
require 'bundler/setup'
require 'sshkit'
require 'sshkit/dsl'

SSHKit.config.output = $stdout

on(:local) do
  execute 'ls -la'
end
```

The output to this looks like this:

    /usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la/usr/bin/env ls -la%

The reason for this is a bit involved, but basically has to do with the fact that `output << cmd` is called for each stdout line, but in this case `output` is `$stdin` rather than an implementation of the `Abstract` formatter. `$stdin` just calls `to_s` on the `command` each time, and the `to_s` evaluates to `/usr/bin/env ls -la`. 

Since this is broken, do you still want to keep support for this syntax? I have a PR in progress to fix this which wraps the specified output in a Formatter if it's not a `SSHKit::Formatter::Abstract` (see below), but maybe, since this is broken and no-one has complained we can just drop support for it. People can always do this:

```ruby
SSHKit.config.output = SomeFormatter.new($stdout)
```
which is covered in the examples.

If we want to keep support for this, we could just format the stream with the `SimpleTextFormatter`

```ruby
def output=(new_output)
  new_output = SSHKit::Formatter::SimpleText.new(new_output) unless new_output.kind_of?(SSHKit::Formatter::Abstract)
  @output = new_output
end
```

In any case I have a PR which cleans up the tests and exposes this issue.

Cheers, Rob

leehambley:
I think it's just broken Rob, that was one of the earlier parts of the code that I wrote, the intention was that you set the output to anything that quacks like an IO, but apparently we've broken that somewhere along the line. I think most people drop in a heavier formatter, so I see the value in making the simple case work, but it's also a bit magic, so we could just remove that misleading line from the docs.

robd:
[...]

PR #244, opened 2015-05-06
Clarify output configuration option

robd:
Currently the docs say we support assigning an io to the `SSHKit.config.output` configuration option:

```ruby
SSHKit.config.output = $some_io
```

Unfortunately this doesn't work at all well as discussed in #243. Therefore, this PR updates the docs to remove that as a supported option. 

It also simplifies some tests which were using the `SSHKit.capture_output` method. This assigned a `String` to `SSHKit.config.output` and then dynamically extended it with a custom write method from `CaptureOutputMethods`, which also added in newlines in some cases. These tests now use the `output` option in the way suggested in the docs, ie:

```ruby
SSHKit.config.output = SSHKit::Formatter::SimpleText.new(@out)
```

This PR also re-enables and fixes the  `simple_netssh` test on the `TestNetssh` functional test, which was being ignored.

Finally, it cleans up some `setup` methods on unit tests which were duplicating the method on `UnitTest` and adds some missing calls to `super`.


robd:
Hey @mattbrictson I don't think this should have any affect on Airbrussh, but I wanted to give you a heads up in case I've missed something.

leehambley:
@mattbrictson I'll wait for an OK from you before merging this, if you'd be kind enough to test before hand!

mattbrictson:
I'll try to review later today (PDT).

mattbrictson:
Looks good to me! I tested this branch and it works fine. :+1:

leehambley:
:busts_in_silhouette: teamwork!

Issue #245, opened 2015-05-06
Proposal - Replace SimpleText formatter with SSHKit.config.log_colors = false option


robd:
@leehambley I didn't want to push a PR on you, but looking at the discussion in #120, you mentioned that you [thought is was bit of a shame to add more formatter logic](https://github.com/capistrano/sshkit/pull/120#issuecomment-37304470) when the SimpleText formatter was added, and I agree. That PR is a bit out of date now that `term-ansicolor` was replaced with `colorize`, but if I understand correctly, the `SimpleText` formatter is just the `Pretty` formatter with colors turned off. 

It seems to me that whether or not you want coloring could be orthogonal to which formatter you use. Certainly, there seems to be a lot of duplication between `SimpleText` and `Pretty`, which makes refactoring in this area harder.

I was wondering if you would be in favour of deprecating the `SimpleText` formatter and introducing a `SSHKit.config.log_colors` (or similar) option.

The implementation for this seems simple - we could just include a check for `SSHKit.config.log_colors` [here](https://github.com/capistrano/sshkit/blob/master/lib/sshkit/color.rb#L24). For the moment, we could keep the `SimpleText` class but delete the duplicated `write` and `write_command` methods. Then we could add a new constructor like this:

```ruby
class SimpleText < Pretty
  def initialize(output)
    SSHKit.config.log_colors = false
    # (I will come up with better deprecation wording here about how to migrate the various options)
    output << 'The SimpleText formatter is now deprecated, and will be deleted soon....'
  end
end
```

This would have the advantage of being able to disable colors on the `Dot` formatter.

Of course I would update the docs, changelog and tests as usual.

leehambley:
> but if I understand correctly, the SimpleText formatter is just the Pretty formatter with colors turned off.

Absolutely

> I was wondering if you would be in favour of deprecating the SimpleText formatter and introducing a SSHKit.config.log_colors (or similar) option.

Maybe respecting `$stdout.tty?` and [`ENV['CLICOLOR']`](http://bixense.com/clicolors/) in some combination (`$stdout.tty?` to set the default value of those env vars?)?

mattbrictson:
Doesn't SSHKit already honor `ENV["SSHKIT_COLOR"]`?

https://github.com/capistrano/sshkit/blob/db4d1d3cb4c95ce40fc28782d68b21858021e2c0/lib/sshkit/color.rb#L24

robd:
[...]

Issue #39, opened 2013-10-28
Capistrano: run with options --dry-run


stamm:
```
cap local deploy --dry-run --trace
** Invoke local (first_time)
** Execute local
** Invoke load:defaults (first_time)
** Execute load:defaults
** Invoke deploy (first_time)
** Execute deploy
** Invoke deploy:starting (first_time)
** Execute deploy:starting
** Invoke deploy:check (first_time)
** Execute deploy:check
** Invoke git:check (first_time)
** Invoke git:wrapper (first_time)
** Execute git:wrapper
cap aborted!
undefined method `verbosity' for "/usr/bin/env #<StringIO:0x007ff7a9827118> /tmp/git-ssh.sh":String
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/formatters/pretty.rb:10:in `write'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/backends/printer.rb:14:in `block in execute'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/backends/printer.rb:13:in `tap'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/backends/printer.rb:13:in `execute'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/capistrano-3.0.0/lib/capistrano/tasks/git.rake:11:in `block (3 levels) in <top (required)>'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/backends/printer.rb:9:in `instance_exec'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/backends/printer.rb:9:in `run'
/Users/stamm/.rbenv/versions/2.0.0-p247/lib/ruby/gems/2.0.0/gems/sshkit-1.1.0/lib/sshkit/runners/parallel.rb:12:in `block (2 levels) in execute'
Tasks: TOP => git:check => git:wrapper
```

[...]

Issue #214, opened 2015-01-15
SSHKit::Runner::ExecuteError: Exception while executing as deploy@myserver.com: undefined method `colorize' for :deploy:Symbol


KurtRMueller:
First, thank you for all the hard work you guys have done on this project and capistrano. Capistrano is a critical a tool in the development, maintenance, and deployment of my rails apps. I don't know if you guys hear that enough, but you should know that almost every rails dev I talk to uses capistrano to deploy. So thank you again.

Now, on to it:

Gem versions:
```
capistrano 3.3.5
sshkit 1.6.1
colorize 0.7.5
```

Whenever I attempt to attempt to deploy via the `cap deploy` task, I get the following error: 
```
SSHKit::Runner::ExecuteError: Exception while executing as deploy@server.com: undefined method `colorize' for :deploy:Symbol
```
After running a stack trace, the error pointed to `/lib/sshkit/color.rb:7`.
```ruby
module Color
  String.colors.each do |color|
    instance_eval <<-RUBY, __FILE__, __LINE__
      def #{color}(string = '')
        string = yield if block_given?
        colorize? ? string.colorize(:color => :#{color}) : string  # LINE 7
      end
    RUBY
  end

  # omitted
end
```

The quick fix for me was simply calling `to_s` on `string`

```ruby
colorize? ? string.to_s.colorize(:color => :#{color}) : string
```

I don't know if this is the right solution (probably not), but it solved my problem. If you like, I can submit a pull request.

leehambley:
Hey @KurtRMueller the problem is `colorize` gem, which recently released a 7.5.0 version, if you check what you have, and try using an old(er) one, it should be fine, we've a couple of open PRs for this already, but there's some more cleanups I wanted to tackle other than just monkey patching the one liner colour issue.

It turns out (in fairness to the Colorize author) that we were using a kind-of internal API of theirs, which finally went away, our bad.

We have a general policy of not *strictly* policing our version dependencies, there'd be hardly anything more annoying than Capistrano (via SSHKit, for example) forcing you to use … for deployment, and only for deployment some arcane combination of Gems which you might very legitimately need at other versions for your *actual* application, so we need to find a solution that works for a broad range of `colorize` versions.

joshuapaling:
For anyone still having the same issue, the really short story is this:
```
  # in your Gemfile
  gem 'capistrano-rails'
  gem 'colorize', '0.7.4'
```

and then run `bundle update colorize`

[...]

PR #246, opened 2015-05-08
No colors for non tty outputs

robd:
As discussed in #245, this PR reworks color support so that we don't ouput ANSI color escape sequences for output IOs which aren't associated with a tty.

The main rework is to convert the `Color` module into a class which can be instantiated with the output IO, rather than just being hard coded to the `$stdout.tty?`

As part of this change, I was trying to add a test for the colors [in this message](https://github.com/capistrano/sshkit/blob/c9de10d38f5ddb81e8bdbacf58ac4e5094277d03/lib/sshkit/formatters/pretty.rb#L13) in the pretty formatter:

```ruby
c.black(c.on_yellow("Output formatter doesn't know how to handle #{obj.class}\n"))
```

Calling write with an unexpected object type didn't call the above line, but instead raised an exception - (`undefined method 'verbosity' for #<Object:0x007f8b73695450>`) [here](https://github.com/capistrano/sshkit/blob/c9de10d38f5ddb81e8bdbacf58ac4e5094277d03/lib/sshkit/formatters/pretty.rb#L8), which meant that the line above was basically unreachable.

I made this line reachable by checking for the presence of the `verbosity` method and I saw this error instead `undefined method 'on_yellow' for Color:Module`. 

I looked through the historical issues on SSHKit and Capistrano about the verbosity and on_yellow errors. There are several verbosity messgage issues which were was associated with the `--dry-run` option and I found [this @leehambley comment](https://github.com/capistrano/sshkit/issues/39#issuecomment-27414697):

> I can confirm the log formatters expect objects on `Command` and `LogMessage` (?) classes. Any thing contrary is incorrect

Therefore, it seems to me that, in practice, we are never calling a formatter with a non `Command` / `LogMessage` object. Therefore, I've added tests to raise this as a more helpful error, and removed the coloring from the message.

Hopefully everything else is obvious.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * Now only color the output if it is associated with a tty, or the `SSHKIT_COLOR` environment variable is set. (See [README](README.md#output-colors)) @robd
   * Removed broken support for assigning an `IO` to the `output` config option (See [#243](https://github.com/capistrano/sshkit/issues/243)). @robd
     * Use `SSHKit.config.output = SSHKit::Formatter::SimpleText.new($stdin)` instead
   * Added support for :interaction_handler option on commands. @robd
--- a/README.md
+++ b/README.md
@@ -359,16 +359,33 @@ However, if you prefer non colored text you can use the `:simpletext` formatter.
 there is also a `:dot` formatter which will simply output red or green dots based on the success or failure of commands.
 There is also a `:blackhole` formatter which does not output anything.
 
-By default, formatters log to `$stdout`, but they can be constructed with any `IO`:
+By default, formatters log to `$stdout`, but they can be constructed with any object which implements `<<`
+for example any `IO` subclass, `String`, `Logger` etc:
 
 ```ruby
-# Output to a StringIO:
-out = StringIO.new
-SSHKit.config.output = SSHKit::Formatter::Pretty.new(out)
-# Do something with out.string
+# Output to a String:
+output = String.new
+SSHKit.config.output = SSHKit::Formatter::Pretty.new(output)
+# Do something with output
 
 # Or output to a file:
 SSHKit.config.output = SSHKit::Formatter::SimpleText.new(File.open('log/deploy.log', 'wb'))
+```
+
+#### Output Colors
+
+By default, SSHKit will color the output using ANSI color escape sequences
+if the output you are using is associated with a terminal device (tty).
+This means that you should see colors if you are writing output to the terminal (the default),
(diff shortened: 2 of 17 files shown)


mattbrictson:
I like this but it has an unfortunate side effect of breaking airbrussh. :sob: 

Airbrussh constructs a pretty formatter with a `Logger` object, which responds to `<<` (and thus works for output purposes). But `Logger` is not an `IO` object and does not respond to `tty?`. And so this happens:

```
NoMethodError: undefined method `tty?' for #<Logger:0x007ff88d10aaa8>
/Users/mbrictson/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/bundler/gems/sshkit-cc6fe03497d4/lib/sshkit/color.rb:15:in `colorize?'
```

Now this is totally my fault, because I am not passing in a proper `IO` object. The reason I am using `Logger` is because of its handy [log rotation feature](http://ruby-doc.org/stdlib-2.2.2/libdoc/logger/rdoc/Logger.html#class-Logger-label-How+to+create+a+logger), which prevents the log file from growing unbounded.

Any ideas how best to solve this?

1. Change airbrussh to use `open` instead of `Logger.new` (i.e. drop the log rotation feature, or roll my own).
2. Change `colorize?` to use a more defensive check: `@io.respond_to?(:tty?) && @io.tty?`.
3. Have airbrussh `extend` the logger object to respond to `tty?`.

robd:
Ah that is a shame - I was hoping the opposite - that this change would stop my `capistrano.log` file being full of color characters!

I think option 2 would be preferable. It narrows the contract for providing an output to SSHKit (ie only `<<` or `write` need to be supported), and people can implement their own `tty?` method if they want more control (ie suggestion 3).

I think it's reasonable that if there is no `tty?` method we default to no color.

People providing a non IO output can wrap with a [SimpleDelegator](http://ruby-doc.org/stdlib-1.9.3/libdoc/delegate/rdoc/SimpleDelegator.html) and add a `tty?` method.

If you're happy with this, I will add tests and update the docs to cover this case.

mattbrictson:
@robd That works for me. Thanks@ :+1:

mattbrictson:
I think this PR might also fix some lingering issues being reported in #214, since the @robd's new `Color` implementation calls `to_s` before invoking `colorize`.

robd:
@mattbrictson OK f37ef77 should fix the `undefined method `tty?` issue

mattbrictson:
> f37ef77 should fix the undefined method 'tty?' issue

Verified!

leehambley:
Good work @robd and thanks again for the compat. sanity checking @mattbrictson - happy to merge if you are @robd

robd:
@leehambley great - yes please!

PR #248, opened 2015-05-11
Remove formatter duplication

### LINKED ISSUE #245: Proposal - Replace SimpleText formatter with SSHKit.config.log_colors = false option

robd:
This PR is a refactor of the `Pretty` / `SimpleText` formatters to remove the duplication between them. I think this is about as dry as it gets for these 2 classes - now SimpleText only has to override the `format_message` in `Pretty` (to exclude the level, uuid and indentation formatting) and the `colorize` method in `Abstract` (to disable coloring). I think this closes #245 which was about having a way of turning off colors in the `SimpleText` formatter, which we have now achieved.

Team @leehambley & @mattbrictson - if you have time to eyeball, would appreciate it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,15 +7,40 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
-  * Now only color the output if it is associated with a tty, or the `SSHKIT_COLOR` environment variable is set. (See [README](README.md#output-colors)) @robd
-  * Removed broken support for assigning an `IO` to the `output` config option (See [#243](https://github.com/capistrano/sshkit/issues/243)). @robd
+  * Simplified formatter hierarchy.
+    [PR #248](https://github.com/capistrano/sshkit/pull/248)
+    @robd
+    * `SimpleText` formatter now extends `Pretty`, rather than duplicating.
+  * Hide ANSI color escape sequences when outputting to a file.
+    [README](README.md#output-colors),
+    [Issue #245](https://github.com/capistrano/sshkit/issues/245),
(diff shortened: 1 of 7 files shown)


mattbrictson:
Looks good! I also did a quick test and didn't see any regressions.

PR #250, opened 2015-05-16
Quote environment variable values

Sinjo:
Ran into an issue this week with an environment variable which contained special shell characters, and broke deploys.

I was going to solve this with `Shellwords.escape`, but found that also escapes '$'.

I ended up taking the approach from #156 and bringing it up to date with master.

@leehambley You okay with this approach?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * Quote environment variable values.
+    [PR #250](https://github.com/capistrano/sshkit/pull/250)
+    @Sinjo - Chris Sinjakli
   * Simplified formatter hierarchy.
     [PR #248](https://github.com/capistrano/sshkit/pull/248)
     @robd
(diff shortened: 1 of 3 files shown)


leehambley:
Yeah, please add an entry to the changelog.

robd:
@Sinjo I had a couple of thoughts about this.

Would it be possible to add a test for the new escaping functionality when double quotes are passed as part of the value?

This is purely a style thing, but I was also wondering if it would be clearer using a `%{}` string literal rather than escaping the double quotes, both in the tests and in `Command`. Maybe this:

```ruby
key_string = key.is_a? Symbol ? key.to_s.upcase : key.to_s
substituted_value = value.to_s.gsub(/"/, '\"')
%{#{key_string}="#{substituted_value}"}
```

Sinjo:
@robd Keen on all those suggestions. I've modified my original commit, and added a couple more to refactor the code/tests.

Look good?

leehambley:
Yep, looks good, particularly fond of the `%{}` for making the tests easier to read.

robd:
@Sinjo That looks great - nice one. @leehambley could probably close #156 too I think?

PR #251, opened 2015-05-17
interaction_handler: Don't put response_data on log

sorah:
interaction_handler is great. but I want keeping use `debug` level logging to get command's output, but when interaction_handler feature puts content of sending data to debug log with message `Sending`.

data sent using interaction_handler may contain sensitive data, so please reconsider logging content by default. Note that the default log level is set to `debug`.

### DIFF
--- a/lib/sshkit/mapping_interaction_handler.rb
+++ b/lib/sshkit/mapping_interaction_handler.rb
@@ -36,7 +36,7 @@ module SSHKit
       if response_data.nil?
         output.debug("Unable to find interaction handler mapping for #{stream_name}: #{data.inspect} so no response was sent")
       else
-        output.debug("Sending #{response_data.inspect}")
+        output.debug("Sending response data")
         if channel.respond_to?(:send_data) # Net SSH Channel
           channel.send_data(response_data)
         elsif channel.respond_to?(:write) # Local IO


leehambley:
Thanks for the patch @sorah, whilst I agree, this will change behaviour and I can't accept it as it. I suggest writing a slightly larger patch which allows the interaction handler to take a callback to be called with the various SSH callbacks (such as `on_response`, etc) which can then be more thoroughly tested, so as not to break the existing behaviour for people

sorah:
okay it's fine.

sorah:
umm I couldn't imagine that acceptable plan. could you implement it instead of me?

robd:
Hi Sorah

Thanks for trying out the interaction handler and giving your feedback! 

I was not sure about logging of the response when I wrote this. I will give this some more thought and try to come up with a PR which allows you to see the output when testing new interaction rules, but disable the output once the rules are finalised.

Thanks,
Rob

sorah:
> try to come up with a PR which allows you to see the output when testing new interaction rules, but disable the output once the rules are finalised.

Yes, that'd be great

robd:
@sorah, @leehambley I think [this approach](https://github.com/capistrano/sshkit/commit/615e5066105327f31b41f228f4b7af46b9b97d32) might work. Basically, no longer log by default, but add support for a log_level parameter when you're first setting up the `InteractionHandler` or trying to debug:

```ruby 
# Initial run
execute(:unfamiliar_command, MappingInteractionHandler.new({}, :debug))
# => DEBUG output: Unable to find interaction handler mapping for stdout: "Some output\n" so no response was sent

# Check log, add mapping
execute(:unfamiliar_command, MappingInteractionHandler.new({
  "Some output\n"=>"Some Input\n"
}, :debug))

# Repeat until mappings are complete
execute(:unfamiliar_command, MappingInteractionHandler.new({
  "Some output\n"=>"Some Input\n",
  "Other output\n"=>"Other Input\n"
}, :debug))

# Once finalised, remove `MappingInteractionHandler` constructor / debug parameter 
execute(:unfamiliar_command, {
  "Some output\n"=>"Some Input\n",
  "Other output\n"=>"Other Input\n"
})
```

If happy, I will create a PR once #249 is merged

robd:
@sorah Does #253 help?

PR #253, opened 2015-05-18
Support log_level constructor parameter for MappingInteractionHandlers

robd:
No longer log by default when using the interaction_handler option in case data is sensitive (passwords etc). Hopefully fixes #251.

### DIFF
--- a/README.md
+++ b/README.md
@@ -336,18 +336,22 @@ execute(:passwd, interaction_handler: lambda { |server_ouput|
 })
 ```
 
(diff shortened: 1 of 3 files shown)


leehambley:
Nice clean patch, thanks again @robd  :)

sorah:
still unconfirmed but I think this work well for us. thanks! :sparkles:

robd:
Great - let me know if you have any more problems or ideas for improvements.

sorah:
just to notify, this patch is working well for us :+1:

PR #256, opened 2015-05-19
Updated CONTRIBUTING guidelines with details on ruby versions, tests, change log and breaking changes

robd:
This is based on my understanding of the process from the last few weeks. Thought this might help people to write PRs which are more likely to be accepted.

Feedback welcome of course!


leehambley:
Excellent work Rob, I miss having this kind of information here. I have a TextExpander macro for suggesting people re-check the Cap CONTRIB docs.

For what it's worth, the download counter https://rubygems.org/gems/sshkit might give you a laugh :) 180,000 "people" using stuff you contributed (extensively) to… maybe a nice little ego (resumé?) boost :)

robd:
:) I'll try not to break anything!

PR #260, opened 2015-05-27
Combine on_stdout and on_stderr into on_data on interaction handler api

robd:
The current `interaction handler` API is overly complex, requiring users to implement both `on_stdout` and `on_stderr`. In many cases, you don't care which stream the data is arriving on. 

Therefore I propose to replace two methods (`on_stdout` and `on_stderr`) with a single `on_data` method. The `stream_name` (`:stdout` or `:stderr`) is now passed as a parameter so can still be used if the user needs to distinguish between the two.

If people are agreed that the single method API is better, we might as well fix this now before the `interaction_handler` API is officially released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -272,8 +272,8 @@ and this can be achieved by specifying an `:interaction_handler` option when you
 **It is not necessary, or desirable to enable `Netssh.config.pty` to use the `interaction_handler` option.
 Only enable `Netssh.config.pty` if the command you are calling won't work without a pty.**
 
-An `interaction_handler` is an object which responds to `on_stdout(stdout, channel, command)` or `on_stderr(stderr, channel, command)`.
-The `interaction_handler`'s methods will be called once per line of `stdout` or `stderr` from the server and
+An `interaction_handler` is an object which responds to `on_data(command, stream_name, data, channel)`.
+The `interaction_handler`'s method will be called once per line of `stdout` or `stderr` from the server and
 can send data back to the server using the `channel` parameter.
 This allows scripting of command interaction by responding to `stdout` or `stderr` lines with any input required.
 
@@ -281,9 +281,9 @@ For example, an interaction handler to change the password of your linux user us
 
 ```ruby
 class PasswdInteractionHandler
-  def on_stderr(channel, stderr, command)
-    puts stderr
-    case stderr
+  def on_data(command, stream_name, data, channel)
+    puts data
+    case data
       when '(current) UNIX password: '
         channel.send_data("old_pw\n")
       when 'Enter new UNIX password: ', 'Retype new UNIX password: '
@@ -373,7 +373,7 @@ execute(:second_command, interaction_handler: ENTER_PASSWORD)
 class PromptUserForPasswordAndCache
   @password_cache = {}
 
-  def on_stderr(channel, stderr, command)
+  def on_data(command, stream_name, data, channel)
     if data =~ /Sorry.*\stry\sagain/
       @password_cache[command.host] = nil
     end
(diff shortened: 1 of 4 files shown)


robd:
@leehambley not sure if your busy time has passed - hope everything is going well with the launch. I wonder what you think about simplifying the `interaction_handler` api? 

If you're happy to merge this, maybe we could release a new sshkit version before I start on any more changes.

robd:
@mattbrictson If you don't mind - can I get your opinion on this change?

mattbrictson:
@robd I don't currently have a need for the interaction handler feature, and it's hard for me to anticipate what the most common use case will be. However I think simpler is almost always better, so this looks good to me.

One possible issue with the interaction handler mechanism in general is that I don't know whether data received from net-ssh necessarily arrives one line at a time. My assumption is that data is delivered as it arrives from the underlying network stack, which depends on network conditions, buffer sizes, etc. For short strings like password prompts it should not be a problem, but it is perhaps something worth noting in the documentation.

Speaking of documentation, I noticed that you gave an example of a stateful handler that caches passwords per host. Starting in capistrano 3.4.0 (I think), the host objects are duplicated before being handed to the backend (https://github.com/capistrano/capistrano/blob/v3.4.0/lib/capistrano/dsl.rb#L53). This means you can't rely on the object id being stable, and so a simple cache using `host` as the hash key won't work. I got bitten by this recently and had to switch to `host.hostname` as the key (https://github.com/mattbrictson/capistrano-mb/commit/a287486e61f8415f5f27c5c0e17b94eda8ce5b6c). In the password cache case, the key might need to be hostname + user?

robd:
@mattbrictson Thanks very much for this feedback. 

Re the common use case, I developed this in order to support interactive sudo - I'm not exactly sure what other use cases people have for this - but I'm assuming cases where capturing input is the only way to make a command work.

Re the data from the server which isn't terminated by a newline - I haven't been able to make this happen as of yet, but I think you are right. I will update the documentation to make this clear. I think in that case, it would make sense to implement an aggregator which combined the partial lines and passed only completed lines to the callback. If I can come up with a test case where I can see the partial lines, I may implement this.

Re the stateful handler example - this is a good catch - thanks. I based this on the [SSHKit Sudo](https://github.com/kentaroi/sshkit-sudo/blob/master/lib/sshkit/sudo/backends_ext/netssh.rb#L31) implementation, but I haven't actually tried it. I'm thinking perhaps I shouldn't include examples unless I've tested them. Perhaps the fact that handlers can be stateful is self explanatory and/or too specific for the readme. Might be better to wait and see what use cases people have and base the docs on what people actually need. I'm happy to amend as you suggest, but maybe I should just delete this example. What do you think?

mattbrictson:
@robd I think you're right: it is probably better just to eliminate the stateful handler example for the time being.

robd:
@leehambley Hey. I saw over at https://github.com/capistrano/capistrano/pull/1465#issuecomment-122193156 that you're thinking about releasing SSHKit today, which is great. I'm not sure if you have had a chance to look at this, but if you are happy, I would be really grateful if we could merge this simplification to the interaction handler API before we release so we don't have to deal with deprecating the 2 method version at a later stage. @mattbrictson recommended some doc changes which I plan to do. I could do these in this PR now as part of this release, or in a separate PR - let me know your preference.

leehambley:
Let's be pragmatic and keep things moving, I'd really appreciate it if I could rely on you to do the doc changes though, as @mattbrictson suggested, so I don't forget to remind you :)

robd:
Great - thanks vm. I'm doing the doc changes right now so I don't forget. I will raise as a separate PR and liaise with @mattbrictson to check the wording.

leehambley:
Alright, awesome :) if not today (well, I am supposed to do *some* client work) it'll be over the weekend come hell or high water :) … so no *huge* rush :) Enjoy your Friday gents, and thanks for your hard work and diligence.

PR #261, opened 2015-06-03
Update EXAMPLES.md

joelataylor:
Added an example of downloading a file from remote to local.


leehambley:
Thanks @joelataylor :)

PR #263, opened 2015-06-24
Remove the need for an external dependency for ANSI color

mattbrictson:
I think SSHKit's ANSI color needs are simple enough that we do not need to depend on a third-party gem (colorize). The colorize dependency has been a source of bugs.

Instead, we could implement a simple module that provides various colorizing methods. These methods could be used as module functions, or mixed in to the class where they are needed for convenience.

I did this for airbrussh, and it has worked well so far:

```ruby
module Airbrussh
  # Very basic support for ANSI color, so that we don't have to rely on
  # any external dependencies.
  module Colors
    ANSI_CODES = {
      :red    => 31,
      :green  => 32,
      :yellow => 33,
      :blue   => 34,
      :gray   => 90
    }.freeze

    module_function

    # Define red, green, blue, etc. methods that return a copy of the
    # String that is wrapped in the corresponding ANSI color escape
    # sequence.
    ANSI_CODES.each do |name, code|
      define_method(name) do |string|
        "\e[0;#{code};49m#{string}\e[0m"
      end
    end
  end
end
```

If there is interest I can create a PR for SSHKit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -56,6 +56,7 @@ appear at the top.
     * Backend implementations now only need to implement `execute_command`, `upload!` and `download!`
     * Removed `Printer` from backend hierarchy for `Local` and `Netssh` backends (they now just extend `Abstract`)
     * Removed unused `Net::SSH:LogLevelShim`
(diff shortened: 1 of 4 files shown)


mattbrictson:
Voilà! Transformed this issue into a PR. Let me know what you think.

leehambley:
Merged :) Thanks @mattbrictson.

Terminal colour code support, and font weighting, etc is a much more complex topic than many realise, and I hope to make time to do a decent integration in SSH kit sometime that actually works correctly.

PR #264, opened 2015-07-14
Escape percentage sign in env variables

Fodoj:
Before:

`ArgumentError: invalid value for Integer(): "/usr/bin/env git fetch origin"` if env variable has percentage sign in it

Now:

works fine.

I guess inside `within` method this issue will still remain if someone will do `in: "/usr/i_love_cra%zy_folders"`, but I assumed that it's rather rare usecase and no-one should have % inside folder names.

### DIFF
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -156,7 +156,7 @@ module SSHKit
     def environment_string
       environment_hash.collect do |key,value|
         key_string = key.is_a?(Symbol) ? key.to_s.upcase : key.to_s
-        escaped_value = value.to_s.gsub(/"/, '\"')
+        escaped_value = value.to_s.gsub(/"/, '\"').gsub("%", "%%")
         %{#{key_string}="#{escaped_value}"}
       end.join(' ')
     end
--- a/test/unit/test_command.rb
+++ b/test/unit/test_command.rb
@@ -49,6 +49,12 @@ module SSHKit
       assert_equal %{( FOO="asdf\\\"hjkl" /usr/bin/env rails server )}, c.to_command
     end
 
+    def test_percent_signs_are_escaped_in_env
+      SSHKit.config = nil
+      c = Command.new(:rails, 'server', env: {foo: 'asdf%hjkl'})
+      assert_equal %{( FOO="asdf%hjkl" /usr/bin/env rails server )}, c.to_command
+    end
+
     def test_including_the_env_doesnt_addressively_escape
       SSHKit.config = nil
       c = Command.new(:rails, 'server', env: {path: '/example:$PATH'})


leehambley:
This topic has been raised before about what escaping, if any to do for environment variables, as there's no way to do it safely and correctly it was decided that we wouldn't implement a half measure.

Fodoj:
@leehambley but there is a half measure already: `.gsub(/"/, '\"')`

leehambley:
Indeed, and one special case concession was already the wrong way to fix it. If you can find a way to reliably escape everything that is required under posix and bash's superset of special tokens. 

* http://www.gnu.org/software/bash/manual/bashref.html#Escape-Character

gives a good reference, but there is no correct implementation in Ruby that we have found.

Sinjo:
Was about to submit my own PR for a similar issue, but noticed this one and thought it would be better to comment here.

In `command.rb`, `sprintf` is used pretty extensively. In a few cases (e.g. [the `with` method](https://github.com/capistrano/sshkit/blob/v1.7.1/lib/sshkit/command.rb#L163), `environment_string` is interpolated into a string which is then used in `sprintf`. If any environment variable contains a '%' symbol, an error ("malformed format string") will be raised as a result.

I've written [a small change](https://github.com/capistrano/sshkit/compare/master...Sinjo:fix-percent-in-env-vars) which fixes the `with` method, and would be happy to extend it to other affected methods.

My thought on why that escaping is okay is that it's not trying to protect the user from the realities of shell interpolation, rather it's making sure valid strings are passed to `sprintf` wherever it's used.

Let me know what you think!

leehambley:
> My thought on why that escaping is okay is that it's not trying to protect the user from the realities of shell interpolation, rather it's making sure valid strings are passed to sprintf wherever it's used.

I'd tend to agree, but there are no absolutes, and environment variables containing `%` are not valid in most shells. I'll look at this more thoroughly when I have time ,thanks @Sinjo

Sinjo:
That's absolutely fair.

Do you think there's an argument for failing early by default if any environment variable contains a '%' character in that case, with an option for people to say "I know which shell I'm using, and it'll be okay"?

I realise that would be a breaking change, so needs more careful planning, and would take longer to make its way to a release, but it feels like it would be nice to fail early and explicitly where possible.

leehambley:
The only internet based citation I can find is http://stackoverflow.com/a/2821183/119669 - But it's highly possible to misread, even the two quoted paragraphs form the open standards document referenced. I'd be fine, as I said, let me review the change a bit more thoroughly when I have change.

Sinjo:
Thanks! Didn't intend to pressure/rush the change at all. Just always keen to improve my knowledge on things. :)

leehambley:
All fine mate, by the way, please update the changelog, etc so I can merge this -

leehambley:
Closing, as we have #280 now

PR #281, opened 2015-09-23
add `:before` option to run `source ~/.bashrc`

sugitak:
# About

This pull request adds `:before` option on `Command`, which let us run other preparation command beforehand.
Typical usage is to run `source ~/.bashrc` before running a command.

# Problem

When I run Capistrano, I always run into the environment variable problem. I just want to read `~/.bashrc` to set my `$PATH`, but I have to set it in my `config/deploy.rb`.

Adding environments variable is not always a good way. Situations I have experienced is:

  - Ruby is installed via `rbenv`, which means I can't run `bundle exec rake` unless I read `~/.bash_profile`
  - Environment variables includes confidential information, so they should only be in the production servers, not in deployment codes or in deploy server

# Usage

Example from the test:

```
irb(main):001:0> puts Command.new(:touch, 'somefile', user: 'bob', env: {a: 'b'}, in: '/var', before: 'source ~/.bashrc').to_command
cd /var && umask 007 && ( export A="b" ; sudo -u bob A="b" -- sh -c 'source ~/.bashrc ; /usr/bin/env touch somefile'
```

### DIFF
--- a/lib/sshkit/backends/abstract.rb
+++ b/lib/sshkit/backends/abstract.rb
@@ -75,6 +75,10 @@ module SSHKit
         remove_instance_variable(:@_env)
       end
 
+      def prerun(command, &_block)
+        (@before ||= []).push command.to_s
+      end
+
       def as(who, &_block)
         if who.is_a? Hash
           @user  = who[:user]  || who["user"]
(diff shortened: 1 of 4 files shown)


sugitak:
Used in pratical situation, there seems to be some problems left... sorry. I should be fixing the problems.

  * prerun commands takes the `default_env`, and the main command can't receive the env
  * want to set `default_prerun`, but running prerun commands before every command may break the behavior

leehambley:
Would this be better solved with the more general solution I proposed in #280 ?

leehambley:
@sugitak you didn't respond to my query 6 months ago about whether the proposed solution in #280 was viable, please report back, and/or close this issue.

PR #285, opened 2015-09-29
add SSHKit::Backend::ConnectionPool#close_connections

akm:
If sshkit is used not in capistrano but in a daemon, SSH connections are kept in CollectionPool. But there is  no method to close them explicitly, so I added SSHKit::Backend::ConnectionPool#close_connections method to close them.

Suppressed net-ssh version `< 3.0.0` to prevent installing error on ruby 1.9.2 and 1.9.3.
- https://travis-ci.org/capistrano/sshkit/jobs/82688897
- https://travis-ci.org/capistrano/sshkit/jobs/82688898

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ appear at the top.
 
 ## `master` (Unreleased)
 
+  * add SSHKit::Backend::ConnectionPool#close_connections
+    [PR #285](https://github.com/capistrano/sshkit/pull/285)
+    @akm
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
   * Clean up rubocop lint warnings
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -33,7 +33,6 @@ module SSHKit
       def close_connections
         @mutex.synchronize do
           @pool.values.flatten.map(&:connection).uniq.each do |conn|
-            $stdout.puts conn.inspect
             if conn.respond_to?(:closed?) && conn.respond_to?(:close)
               conn.close unless conn.closed?
             end
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -30,6 +30,18 @@ module SSHKit
         end
       end
 
+      def close_connections
+        @mutex.synchronize do
+          @pool.values.flatten.map(&:connection).uniq.each do |conn|
+            $stdout.puts conn.inspect
+            if conn.respond_to?(:closed?) && conn.respond_to?(:close)
+              conn.close unless conn.closed?
+            end
+          end
+          @pool.clear
+        end
+      end
+
       def flush_connections
         @mutex.synchronize { @pool.clear }
       end
--- a/sshkit.gemspec
+++ b/sshkit.gemspec
@@ -17,7 +17,7 @@ Gem::Specification.new do |gem|
   gem.require_paths = ["lib"]
   gem.version       = SSHKit::VERSION
 
-  gem.add_runtime_dependency('net-ssh',  '>= 2.8.0')
+  gem.add_runtime_dependency('net-ssh',  '>= 2.8.0', '< 3.0.0')
   gem.add_runtime_dependency('net-scp',  '>= 1.1.2')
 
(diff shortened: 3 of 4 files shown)


akm:
@ppp0 I modified the dependency to use '~> 2.8' and it works. Thank you.

ppp0:
Excellent, I'm glad, thank you very much!

leehambley:
That seems lke a strange way to fix things, we don't want to get stuck on obsolete Net-ssh, I would have been happy to merge the original, but with this locking to an old version, I can't really. Happy to see the tests having passed.

ppp0:
@leehambley As far as I understand `~> 2.8` will install up to `2.9.x` the last version requiring ruby 1.9 - Obsolete in what sense? `3.x` is a major step forward and requires `ruby >=2.0` which breaks my capistrano install and very likely a lot more gems that depend on ssh-kit.

To mitigate this, it would probably be good to tag a major version for ssh-kit, too, so that dependent libraries can reliably pin the version.

leehambley:
Thanks we should probably force `~> 2.8.x` then, where `x` is whatever was most recently released. I saw that Net::SSH was making a move towards requiring Ruby 2.0, I could imagine us doing the same, I'd need to check what we do about Net::SSH, and what if any API changes over there warrant a forced minimum Ruby version.

ppp0:
I personally would stick with `~> 2.8` as it covers all releases up to `2.9.x`
If you look at https://github.com/net-ssh/net-ssh/releases, there is only one `2.8.0` release. `3.x` is the breaking point, ie when ruby >=2.0 is required

akm:
hi @leehambley

 I think there are 6 options about this:

dependency       | ruby-1.9.2,1.9.3,2.0.0,2.1.2.2.2 | ruby-2.0.0,2.1.2.2.2
-----------------|----------------------------------|----------------------------
net-ssh >= 2.8.0 | [A] Failure because net-ssh-3.0.1 requires ruby >= 2.0.0 | [D] Success
net-ssh ~> 2.8.x | [B] Success (*1)                  | [E] Success (*1)
net-ssh ~> 2.8   | [C] Success                       | [F] Success

*1 It doesn't support net-ssh-2.9.x and 2.10.x according to http://guides.rubygems.org/patterns/#declaring-dependencies

I recommend [C] because it is the most extensive support for ruby and net-ssh.

mattbrictson:
Ruby 1.9.3 has been removed from the Travis matrix, so the original form of this PR (without the net-ssh pin) should now pass CI checks.

@akm Perhaps you could rework this PR to remove the gemspec change?

ppp0:
@mattbrictson apologies, I was asking to pin the version because without, I will have a couple of things depending upon this that will break. Is it an unavoidable technical need to require ruby 2 here? Don't tests pass with versions pinned for ruby 2, too?

akm:
@mattbrictson I did. Is this alright?

ppp0:
>support for ruby 1.9 was dropped
:cry:

leehambley:
> support for ruby 1.9 was dropped

To clarify Net-SSH 3.x drops support for 1.9, and since we don't want to force people to use an older version of net-ssh (security, and our general policy on being liberal-to-a-fault with our dependency version requirements) - if *you* choose to lock Net-SSH at 2.x in your Gemfile, Cap and SSHKit will continue to work on 1.9 for the for seeable future. (at least 12 months past the [Ruby 1.9 EOL](https://www.ruby-lang.org/en/news/2014/01/10/ruby-1-9-3-will-end-on-2015/)). I'll document some of this in our README, or at least in the release notes.

Frankly we'd love to be able to use [keyword arguments](https://robots.thoughtbot.com/ruby-2-keyword-arguments), and [`Module#prepend`](http://gshutler.com/2013/04/ruby-2-module-prepend/) amongst other things. So eventually 1.9 support will have to go, but it's more a question of practicality.

mattbrictson:
Also, the fact that Bundler resolves net-ssh to 3.x on Ruby 1.9 (thereby leading to an error) is due to a limitation of the current Bundler/rubygems. If all goes well [it will be fixed soon](https://github.com/bundler/bundler/issues/4046).

leehambley:
Thanks for linking that issue Matt, so I feel vindicated in dropping the
1.9 suite, even if we don't explicitly require 1.9 yet. I'll add a note to
the SSHKIt and cap readmes

PR #288, opened 2015-10-15
Avoid the usage of a double negative.

will-in-wi:
I was going through the code and had to read this message a couple of times. I hope that this wording is clearer.

Thanks for all of your work on this!


will-in-wi:
The build issue looks like something is wrong with Travis CI unrelated to the change. Could someone kick it?

PR #289, opened 2015-10-22
Monkey patch net-ssh to make hashing scheme work

dalehamel:
@leehambley @mattbrictson for review

[this line](https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/connection_pool.rb#L19) causes sshkit to call inspect on "new_connection_args" and stringify it. This causes the **memory address** of the proxy object to be included in the hash key, meaning that identical objects that just have a different address won't get a hash hit. This causes it to generate new connections over and over and over as it will never find the connection it's looking for, so it just generates new ones and shoves them in the hash.

For instance:

This key:
```
["SOME_HOST", "deploy", {:user=>"deploy", :forward_agent=>true, :proxy=>#<Net::SSH::Proxy::Command:0x00000000b8c9a0 @command_line_template="ssh deploy@SOME_BASTION nc %h %p", @command_line=nil>}]
```

Doesn't match this key:
```
["SOME_HOST", "deploy", {:user=>"deploy", :forward_agent=>true, :proxy=>#<Net::SSH::Proxy::Command:0x000000010010a8 @command_line_template="ssh deploy@SOME_BASTION nc %h %p", @command_line=nil>}]
```

Because the memory address is different, even though this connection could totally be reused, so the code bails out of find_live_entry https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/connection_pool.rb#L41

The patch here is actually to support 'inspect' in sshkit, and use the command line template, which is all way care about. This will cause the hashes to properly hit again, and stops the run away connections.

In our case, we have 2 gateways we need to proxy through - one for each region in amazon that our app lives in. For this reason, we specify the proxy **on each host** so that the use the correct proxy (the one in their region). This causes the code to render a new proxy command object for each host, on each command.

In our case, with just 12 hosts in each region, essentially every command would need a new connection. This caused an SSH command to run for each capistrano command, and these commands would never be reaped. The pipe used by each command, along with the shared objects, caused the file descriptors used by capistrano and it's children to easily blow past the default of 1024. Not to mention substantially slowing down the deploy.

Under this scheme, we can again just have a single proxy for each host which is a much more manageable number of proxies.

**This is definitely not a pretty solution**, I would prefer just generating the key in a different way, but this is the most elegant thing I could think of given the current structure.

I would happily use a different hashing scheme, so long as it's consistent and based on the object attributes, and not the object addresses.

Solves the issue i had with proxy command leaking processes and FDs mentioned in https://github.com/capistrano/capistrano/issues/725

### DIFF
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -1,5 +1,19 @@
 require "thread"
 
+# Since we call to_s on new_connection_args and use that as a hash
+# We need to make sure the memory address of the object is not used as part of the key
+# Otherwise identical objects with different memory address won't get a hash hit.
+# In the case of proxy commands, this can lead to proxy processes leaking
+# And in severe cases can cause deploys to fail due to default file descriptor limits
+# An alternate solution would be to use a different means of generating hash keys
+module Net; module SSH; module Proxy
+  class Command
+    def inspect
+      @command_line_template
+    end
+  end
+end;end;end
+
 module SSHKit
 
   module Backend


dalehamel:
It might actually be the case that any object that's passed in as an SSH option that doesn't have a simple inspect or to_s could have the same problem.

A proxy command is probably going to be the most likely one, but for that reason it's probably better to more explicitly define what the hash key should be, rather than try and catch all cases via monkey patching

leehambley:
@mattbrictson at least this is a neat case of "cache key miss", so I hope we can unit test it now that @dalehamel has been kind enough to send a mostly-viable fix.

I'd be uncomfortable merging it as it is, but since you and I are as busy as we are, maybe @dalehamel can use his SSHKit fork with the `github: '.../...'` trick, and we can solve this "properly" soon, this issue is visible enough, and I don't think the world will come to an end if we make for a more robust lookup.

Thanks for the patch @dalehamel this kind of stuff can be a nightmare, I know, and your contribution makes it about 9000% more likely we'll be able to fix it :)

leehambley:
> A proxy command is probably going to be the most likely one, but for that reason it's probably better to more explicitly define what the hash key should be, rather than try and catch all cases via monkey patching

Indeed, I see by the way you fixed it that we should maybe come up with a better scheme for IDing things uniquely enoguh to handle this

dalehamel:
@leehambley I we can use our fork for now I suppose, but we have several projects that will need this fix so it'll be a bit of trouble to have to track them down and fix them all.

Perhaps we could merge this to make it into the upcoming release, and I'll draw up another PR that attempts to fix the problem properly and more generally.

As it is though, this is a strict improvement, I can't see this breaking any functionality. I wonder how many people are having the same problem but not realizing? This could speed up a lot of deploys and make them more reliable.

leehambley:
> As it is though, this is a strict improvement, I can't see this breaking any functionality. I wonder how many people are having the same problem but not realizing? This could speed up a lot of deploys and make them more reliable.

True enough, @mattbrictson  knows much more about this pooling code than I and even @robd knows a little.

I can't *really* say with any degree of confidence what the side effects might be, but it does look like `key.to_s`  was a bit of an arbitrary decision and probably would be harmless to change it.

leehambley:
It's funny, I'm noticing that this line came from a PR claiming to fix pooling across threads - and none of the regular contributors or maintainers were really in the loop - https://github.com/capistrano/sshkit/commit/3c01aaefdcc55be41b0a1ad8dd7d36dcd5112a0f

dalehamel:
Yeah I came to the same conclusion about that PR.

mattbrictson:
@dalehamel Is this a correct understanding of how you have things set up? (pseudo-code):

```
server "region1-app01", proxy("region1-gateway")
server "region1-app02", proxy("region1-gateway")
server "region1-app03", proxy("region1-gateway")
server "region2-app04", proxy("region2-gateway")
server "region2-app05", proxy("region2-gateway")
server "region2-app06", proxy("region2-gateway")
```

If so, the connection pool will contain 6 distinct connections, right?

Even if we monkey patch `inspect` for the proxy object, we'll still end up with 6 connections, because there are still 6 distinct hostnames.

Further, I don't understand how you are seeing a new proxy object memory address *on each command execution*. The proxy is constructed once, in `deploy/production.rb`, right? Is there a `.dup` being called somewhere? How are the new proxy objects being constructed on each command execution?

I am thoroughly confused!

*Edit: OK, I see that a `.dup` (in the form of `Marshal`) snuck in here: https://github.com/capistrano/capistrano/commit/837eaca8b1585c414ddc0718a55fe0f592b3c79a. That explains it.*

dalehamel:
Yeah your pseudo code is basically correct, and the Marshall explains the
new proxies.

Are you still confused?

mattbrictson:
> Are you still confused?

Nope, I got it now. Although I am not sure how we can generate an appropriate cache key without using the monkey patch you've supplied. Since the host object is duplicated, nothing is stable, not even the memory addresses. I think your solution may be the best we can do, unless we roll back the commit that introduced the `Marshal` stuff.

dalehamel:
I guess we could flatten the attributes of all nested objects in some way,
such that we cut out the memor addresses?

I think you're probably right the, the monkey patch is the simplest thing I
can think of that covers the majority of cases without introducing a major
change.

dalehamel:
@mattbrictson @leehambley given that this seems like the best we can do, can we get this merged?

The CI failures seem to be old versions of ruby, which don't seem to be supported at all anyways?

leehambley:
We're considering dropping support for 1.9 anyway. 

@mattbrictson I'm going to merge this, and open an issue in it's stead.

PR #292, opened 2015-10-26
Add chandler to `rake release`; improve docs

mattbrictson:
Chandler is a small gem that plugs into Bundler's `rake release` rake task. Upon successful release to rubygems.org, Chandler will parse the SSHKit CHANGELOG and upload the release notes it finds there to the Releases page on GitHub.

This ensures that the GitHub releases page always has our latest release notes, without requiring tedious copy and paste.

As an example, if you look at the [Releases](https://github.com/capistrano/sshkit/releases) page now, you can see the 1.7.1 release notes. These were parsed from the existing CHANGELOG and uploaded by running `chandler push v1.7.1`. This PR makes this process automatic by doing a `chandler push` whenever we `rake release`.

I made changes to the `RELEASING.md` to explain the prerequisites for using `rake release`. Notably, chandler requires GitHub API access, which means you need to configure GitHub in `~/.netrc`.

Note that since Chandler requires Ruby 2.1+, and SSHKit supports back to 2.0, I've added some conditional logic in the Gemfile and Rakefile to ensure the build does not blow up in a Ruby 2.0 environment.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,3 +6,10 @@ platforms :rbx do
   gem 'rubysl', '~> 2.0'
   gem 'json'
 end
+
+# Chandler requires Ruby >= 2.1.0, but depending on the Travis environment,
+# we may not meet that requirement. Only include the chandler gem if the Ruby
+# requirement is met. (Chandler is used only for `rake release`; see Rakefile.)
+if Gem::Requirement.new('>= 2.1.0').satisfied_by?(Gem::Version.new(RUBY_VERSION))
+  gem 'chandler', '>= 0.1.1'
+end
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -1,10 +1,18 @@
 # Releasing
 
-* **Ensure the tests are passing.**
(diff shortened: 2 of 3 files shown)


leehambley:
> Note that since Chandler requires Ruby 2.1+, and SSHKit supports back to 2.0

Might I ask what warranted using 2.1+ only features (what feature did I miss from 2.1?)

mattbrictson:
Refinements were considered experimental in 2.0, and not officially released until 2.1. That's the main reason for the 2.1 requirement.

leehambley:
> Refinements were considered experimental in 2.0, and not officially released until 2.1. That's the main reason for the 2.1 requirement.

Cheers - did the APIs change, or was it just a case of it being less than final, so you were being cautious?

mattbrictson:
I believe the main difference is that refinements couldn't be applied within a Module until 2.1. I don't remember if I use that feature specifically in Chandler, but yes, I wanted to err on the side of caution.

Come to think of it, I'm also using required keyword arguments, which weren't introduced until 2.1.

PR #293, opened 2015-11-01
Use env vars instead of Etc module for local user lookup

felixbuenemann:
This ensures that the code works on both Unix and Windows, while `Etc.getpwuid` relied on /etc/passwd being present, so it returned a NoMethodError when calling `name` on Windows.

This change has been discussed in capistrano/capistrano#1521.


felixbuenemann:
@leehambley ping.

leehambley:
Cheers!

PR #294, opened 2015-11-01
New API for setting+configuring formatter: use_format

mattbrictson:
As discussed in https://github.com/mattbrictson/airbrussh/pull/64#issuecomment-152255622, this PR adds a new API for setting the SSHKit output.

Before:

```ruby
SSHKit.config.format = :pretty
```

After:

```ruby
SSHKit.config.use_format :pretty
```

The benefit is that the new `use_format` accepts arbitrary arguments that are then passed to the formatter's constructor. This allows a formatter to be configured.

Hypothetical example:

```ruby
SSHKit.config.use_format :airbrussh, :color => true, :truncate => false
```

This would construct `SSHKit::Formatter::Airbrussh` and pass `{ :color => true, :truncate => false }` to its constructor.

SSHKit's current formatters do not yet accept extra arguments, but this API lays the groundwork for future formatters or formatter gems like airbrussh that are configurable and would benefit from a standard API for doing so.

Eventually we could deprecate the `format=` API, but that can be a separate PR.

@leehambley, @robd Let me know if this looks good and I'll expand this PR to a CHANGELOG entry and update the docs.

Also: this PR fixes a bug where calling `format = :bad_value` proceeds with no error or warning, but blows up later once a logging operation is attempted. Now it will immediately raise `NameError`.

### DIFF
--- a/lib/sshkit/configuration.rb
+++ b/lib/sshkit/configuration.rb
@@ -53,9 +53,16 @@ module SSHKit
     end
 
     def formatter(format)
-      SSHKit::Formatter.constants.each do |const|
-        return SSHKit::Formatter.const_get(const).new($stdout) if const.downcase.eql?(format.downcase)
+      formatter_class(format).new($stdout)
+    end
+
+    def formatter_class(symbol)
+      name = symbol.to_s.downcase
+      found = SSHKit::Formatter.constants.find do |const|
+        const.to_s.downcase == name
       end
+      fail NameError, 'Unrecognized SSHKit::Formatter "#{symbol}"' if found.nil?
+      SSHKit::Formatter.const_get(found)
     end
 
   end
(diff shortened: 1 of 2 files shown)


robd:
:+1: This looks good to me

leehambley:
Thanks @mattbrictson nice work, sorry merged out of habit, and then reverted. I'm around for an hour more thing evening if you have time, looks good to me though.

mattbrictson:
@leehambley It should be good to merge now.

I updated the README to refer to `use_format`, and I also **removed** references to using a `String` or `File` as a formatter (i.e. anything that has a `<<` method), because those no longer work with SSHKit's new fine-grained logging methods (e.g. `log_command_start`).

I also put entries in the CHANGELOG.

mattbrictson:
Hmm.. GitHub is not pulling the new commits into the PR. Do I need to create a new PR?

mattbrictson:
@leehambley Please merge #295 instead. Thanks!

PR #296, opened 2015-11-01
New API for setting+configuring formatter: use_format

mattbrictson:
Redo of #294 to include README and CHANGELOG updates.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -76,6 +76,8 @@ appear at the top.
     * Removed `Printer` from backend hierarchy for `Local` and `Netssh` backends (they now just extend `Abstract`)
     * Removed unused `Net::SSH:LogLevelShim`
   * Removed dependency on the `colorize` gem. SSHKit now implements its own ANSI color logic, with no external dependencies. Note that SSHKit now only supports the `:bold` or plain modes. Other modes will be gracefully ignored. [#263](https://github.com/capistrano/sshkit/issues/263)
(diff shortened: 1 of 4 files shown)


mattbrictson:
@leehambley I've rebased the branch so that you merge this into master and replay the commits that were reverted.

Should be good to merge :+1:

robd:
@mattbrictson I thought I had reinstated support for any output which implemented `<<`. Are you sure  this isn't implemented any more?

leehambley:
Thanks @mattbrictson :) release due Nov 02 :)

mattbrictson:
@robd You're right. I was skimming the README and thought it was referring to `SSHKit.config.output = $stdout`, when in fact it was referring to passing `$stdout` to the formatter. I'll fix it.

robd:
@mattbrictson There is an argument to say perhaps we could get rid of the output config option and just allow the use_format option to take an instantiated `Formatter`:

```ruby
output = String.new		
SSHKit.config.use_format(SSHKit::Formatter::Pretty.new(output))
```

Not sure if this is better?

Also, in the docs, should it be this:

```ruby
SSHKit.config.use_format :foobar
...
SSHKit.config.use_format :foobar, :my_option => "value"
``` 
instead of 

```ruby
SSHKit.config.use_format = :foobar
...
SSHKit.config.use_format = :foobar, :my_option => "value"
```

mattbrictson:
> Also, in the docs, should it be this

Yes I noticed that too. Fixed in #297. (My commits have been pretty sloppy today, sorry!)

robd:
No problem - I saw you'd spotted too after I commented - thanks for sorting these.

PR #297, opened 2015-11-01
Reinstate output doc I deleted by mistake

mattbrictson:
Reinstate the doc I accidentally removed in https://github.com/capistrano/sshkit/pull/296#issuecomment-152871755

Good catch, @robd. Sorry about that!


leehambley:
I assumed nuking that was intentional, but I suppose it does still work.

PR #298, opened 2015-11-04
Close expired pool connections

ziuchkovski:
Ensure expired pool connections are closed.

I encountered this issue using sshkit with a large number of hosts.  My setup relies on jump hosts. `Net::SSH` is correctly parsing the jump host config from my ~/.ssh/config, and is connecting just fine.  However, since sshkit isn't closing the expired connections, I'm accumulating hundreds/thousands of external processes (the result of `ProxyCommand` in my ~/.ssh/config).  Eventually this starves my system of resources and the ruby process dies.

Anyway, this PR ensures expired connections are closed and pruned at checkout and checkin.  Ultimately the pool might make more sense as an LRU cache with an upper bound on number of entries, but that seemed like too much change at once.

### DIFF
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -29,20 +29,23 @@ module SSHKit
       end
 
       def checkout(*new_connection_args, &block)
-        prune_expired
-        # Optimization: completely bypass the pool if idle_timeout is zero.
+        entry = nil
         key = new_connection_args.to_s
-        return create_new_entry(new_connection_args, key, &block) if idle_timeout == 0
-
-        find_live_entry(key) || create_new_entry(new_connection_args, key, &block)
+        if idle_timeout
+          prune_expired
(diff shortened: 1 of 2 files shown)


leehambley:
We would also need you to actively consent to relicensing your contribution under MIT/BSD (we're not sure yet) licences, as it is we're GPL but that is bound to change in the next days.

ziuchkovski:
Absolutely.  I will gladly relicense the contribution under MIT/BSD, or any other license of your choosing.  Anything in particular you'd like me to do, or just grant my consent here?

leehambley:
Consent here is fine :)

ziuchkovski:
@leehambley I pushed one final commit seconds before merge.  Just wanted to make sure you're aware of the change.  The timing on that was uncanny.

leehambley:
Still looks fine! Thanks

ziuchkovski:
Whew, good.  And thanks for merging!

PR #301, opened 2015-11-23
Fix quoting

eagletmt:


leehambley:
Thanks, please update the changelog too if you want to be credited, otherwise you won't be!

eagletmt:
Added CHANGELOG entry referring to this PR.

mattbrictson:
Good catch. Thanks!

PR #304, opened 2015-12-09
Formatters::Pretty: display command being run

steved:
It seemed odd to see the `/usr/bin/env` prefix when that wasn't the command being run.

### DIFF
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -206,7 +206,11 @@ module SSHKit
     end
 
     def to_s
-      [SSHKit.config.command_map[command.to_sym], *Array(args)].join(' ')
+      if should_map?
+        [SSHKit.config.command_map[command.to_sym], *Array(args)].join(' ')
+      else
+        command.to_s
+      end
     end
 
     private
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ appear at the top.
   * make sure working directory for commands is properly cleared after `within` blocks
     [PR #307](https://github.com/capistrano/sshkit/pull/307)
     @steved
+  * display more accurate string for commands with spaces being output in `Formatter::Pretty`
+    [PR #304](https://github.com/capistrano/sshkit/pull/304)
+    @steved
 
 ## 1.8.1
 
--- a/test/unit/backends/test_printer.rb
+++ b/test/unit/backends/test_printer.rb
@@ -22,7 +22,7 @@ module SSHKit
       def test_execute
         printer.execute 'uname -a'
         assert_output_lines(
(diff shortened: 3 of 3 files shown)


robd:
I just wanted to make sure I understand exactly what the problem here is. Eye balling the code, it looks like this bug is that commands which don't use the command mapper (ie ones with spaces in), are incorrectly logged with the command mapper applied. For commands which do use the command mapper, then the log line is correct. Have I understood this right?

One thing I wanted to check, for a command with a lot of options applied, would this change make the log line very long (ie commands which [are wrapped](https://github.com/capistrano/sshkit/blob/e9d520fe40998e86c34f098ec2e096d7cd39e096/lib/sshkit/command.rb#L193-L197) with `umask`, `within`, `user` etc would now print out all of the wrapping details).  From the looks of the code, the intention is not to print all this stuff out for commands using the command mapper. I haven't tried this so I'm not sure if I misunderstood something.

steved:
Ah, I see what you're saying. Would this be ok? I know it isn't DRY, but the methods are so close anyway:

```diff
diff --git a/lib/sshkit/command.rb b/lib/sshkit/command.rb
index d3787d3..cdf9811 100644
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -206,7 +206,11 @@ module SSHKit
     end

     def to_s
-      [SSHKit.config.command_map[command.to_sym], *Array(args)].join(' ')
+      if should_map?
+        [SSHKit.config.command_map[command.to_sym], *Array(args)].join(' ')
+      else
+        command
+      end
     end
```

lib/sshkit/command.rb
```diff
@@ -206,7 +206,11 @@ def to_command
     end
 
     def to_s
-      [SSHKit.config.command_map[command.to_sym], *Array(args)].join(' ')
+      if should_map?
+        [SSHKit.config.command_map[command.to_sym], *Array(args)].join(' ')
+      else
+        command
```
mattbrictson: I think this needs to be `command.to_s`, because `command` may be a symbol.

steved: Yep, done

steved:
bump

mattbrictson:
This looks good, but I haven't had time to think through the consequences. My guess is that `Command#to_s` is used a lot, either explicitly or within string interpolation. Have you tried doing a cap deploy against this branch?

steved:
Yep, looks alright to me. Unfortunately, I'm new to v3 so I'm not 100% sure what the implications are either.

mattbrictson:
OK, I've now tested it as well. Seems good to me! :+1:

steved:
@leehambley

steved:
Thanks! Do you mind releasing a new version?

leehambley:
Next week proobably

Issue #265, opened 2015-07-17
within release_path do; execute "tar zxf /tmp/test.gz"; end appears to untar test.gz in $HOME


vrsean:
This behavior seems counter-intuitive given `within`. The problem was resolved with setting the -C argument to release_path for tar but stumped me a bit.  `execute "pwd"` indicated the proper directly, oddly enough.

The code was executed within the context of `on roles(:assets) do |host|`...basically we are speeding up asset deployment by forking off the compilation of assets in the beginning, tarring up public/assets, and scping that file to /tmp in another task to give some background on the issue.  

We are using sshkit 1.5.1.

Can anyone duplicate this issue or shed some light on it?

Thanks,

--sean

leehambley:
Please re-read [this](https://github.com/capistrano/sshkit#running-commands):

> Note: In SSHKit, the first parameter of the execute / test / capture methods has a special significance. If the first parameter isn't a Symbol, SSHKit assumes that you want to execute the raw command and the as / within / with methods, SSHKit.config.umask and the command map have no effect.

That means you must call:

``` ruby
within release_path do
  execute :tar, '-zxf', '/tmp/test.gz'
end 
```

Not:

``` ruby
within release_path do
  execute "tar zxf /tmp/test.gz"
end 
```

PR #307, opened 2015-12-17
pwd is an empty array if already instantiated

steved:
This might be what was reported in #265. Once a `within` is called and exited, any other mapped commands will execute in the $HOME directory. Still working on tests.

### DIFF
--- a/lib/sshkit/backends/abstract.rb
+++ b/lib/sshkit/backends/abstract.rb
@@ -121,8 +121,16 @@ module SSHKit
         command(args, options).tap { |cmd| execute_command(cmd) }
       end
 
+      def pwd_path
+        if @pwd.nil? || @pwd.empty?
+          nil
+        else
+          File.join(@pwd)
+        end
+      end
+
       def command(args, options)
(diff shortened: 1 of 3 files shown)


steved:
@robd @mattbrictson @leehambley

lib/sshkit/backends/abstract.rb
```diff
@@ -122,7 +122,8 @@ def create_command_and_execute(args, options)
       end
 
       def command(args, options)
-        SSHKit::Command.new(*[*args, options.merge({in: @pwd.nil? ? nil : File.join(@pwd), env: @env, host: @host, user: @user, group: @group})])
+        pwd = (!@pwd || @pwd.empty?) ? nil : File.join(@pwd)
```
mattbrictson: Can you factor this out into a private helper method, e.g. `pwd_path`? Also, `@pwd.nil?` is more readable than `!@pwd` IMHO.

mattbrictson:
Great discovery and good test! I made some minor suggestions for coding style, but otherwise LGTM. I'll let @leehambley have the final say.

lib/sshkit/backends/abstract.rb
steved: Sure, fixed.

leehambley:
Strange that we never ran into the issue.

PR #309, opened 2015-12-26
Upgrade to minitest 5

mattbrictson:
SSHKit was using the `turn` gem, which is no longer maintained. SSHKit was also pinned to a very old version of the minitest gem (released in 2012).

This commit brings SSHKit up to the current 5.x version of minitest, and replaces the turn gem with minitest-reporters, which provides similar rspec-like output with ansi color.

Other changes in minitest 5:

* `MiniTest::Unit::TestCase` is now `Minitest::Test`
* `require 'minitest/autorun'` instead of `MiniTest::Unit.autorun`

Old test output:

```
SSHKit::Backend::TestNetssh
     PASS (0:00:00.214) test_netssh_ext
     PASS (0:00:00.215) test_net_ssh_configuration_options
     PASS (0:00:00.216) test_transfer_summarizer

```

New test output:

```
SSHKit::Backend::TestNetssh
  test_net_ssh_configuration_options                              PASS (0.00s)
  test_transfer_summarizer                                        PASS (0.00s)
  test_netssh_ext                                                 PASS (0.00s)
```

### DIFF
--- a/sshkit.gemspec
+++ b/sshkit.gemspec
@@ -20,9 +20,9 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('net-ssh',  '>= 2.8.0')
   gem.add_runtime_dependency('net-scp',  '>= 1.1.2')
 
-  gem.add_development_dependency('minitest', ['>= 2.11.3', '< 2.12.0'])
+  gem.add_development_dependency('minitest', '>= 5.0.0')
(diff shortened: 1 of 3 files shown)


leehambley:
Good one @mattbrictson, and respect to the MiniTest team that we can jump 3.5 years of development with such minimal changes.

PR #312, opened 2015-12-29
`SSHKit::Backend::Printer#test` now always returns true

mikz:
which makes --dry-run pass the `deploy:check:linked_files`


leehambley:
Ohh dear, `--dry-run` is awful :( are people really relying on that ?

mikz:
@leehambley yes, it is first level sanity check

PR #313, opened 2015-12-31
Add `rubocop --lint` to default rake task and Travis build

mattbrictson:
I see there is a [precedent](https://github.com/capistrano/sshkit/pull/275) for running RuboCop's lint checks on the SSHKit codebase, so I thought we should make it an official part of the build process.

This PR does the following:

* Fixes all warnings reported by `rubocop --lint` (there were only a few)
* Explains our usage of RuboCop in CONTRIBUTING
* Adds a `lint` rake task for executing `rubocop --lint`
* Makes linting part of the default rake task
* Adds linting to the Travis build

RuboCop's linting rules are not very controversial (I hope!), so I figure this is a good start. We can perhaps add stricter style guidelines in the future if there is interest.

/cc @leehambley @robd @cshaffer

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ rvm:
   - 2.2.2
   - 2.1.6
   - 2.0.0
-script: "rake test:units"
+script: "rake test:units lint"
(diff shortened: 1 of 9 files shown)


robd:
:+1: Looks good to me

PR #318, opened 2016-01-22
Ability to specify custom runner

kirs:
This patch enables ability to specify my own runner class.

```ruby
module SSHKit
  module Runner
    class MyParallel < Abstract
    # ...
    end
  end
end

on roles(:web), in: :my_parallel do
  # …
end
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,8 @@ appear at the top.
     helper methods and still have easy access to the currently-executing Backend
     without having to use global variables.
     [PR #319](https://github.com/capistrano/sshkit/pull/319) @mattbrictson
+  * Add `SSHKit.config.default_runner` options that allows to override default command runner.
+    This option also accepts a name of the custom runner class.
 
 ## 1.8.1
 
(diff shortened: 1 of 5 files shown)


mattbrictson:
I have no strong feelings on this one either way. @leehambley Any objections?

Before a merge we would also need:

* CHANGELOG entry
* Unit test
* Rebase against current `master` so that Travis checks pass

leehambley:
I'd rather it took a class name honestly.

kirs:
The idea about class name sounds good. I'll change it. Basically, using custom runner is a use-case from our setup at @Shopify.

leehambley:
I thought you were working at Giant Martians ?

kirs:
I've reworked it a bit:
- `in:` accepts a class name
- you can globally set the default runner with `SSHKit.config.default_runner` (which is still `parallel` by default)

WDYT? I'll add the changelog and tests if we're good with implementation.
@leehambley it's Shopify since 2016 :boom:

leehambley:
Go ahead :)

kirs:
Done. Please have a look on tests :)

leehambley:
@kirs Travis blew up mate, could you take a look?

kirs:
@leehambley fixed.

PR #322, opened 2016-01-29
Fix race condition in CommandMap#prefix

mattbrictson:
`PrefixProvider#[]` could return `nil` in multi-threaded scenarios due to a race condition involving `||=`, which is not an atomic operation in JRuby.

The test that I added failed with JRuby 9.0.4.0 before the fix; now it passes.

Fixes https://github.com/capistrano/capistrano/issues/1528


leehambley:
Excellent catch @mattbrictson

leehambley:
(I know you wanted some feedback from the jRuby guys who were struggling, but the patch is harmless)

PR #324, opened 2016-01-30
Use Refinements instead of core ext and do not pollute the global scope

kirs:


kirs:
Ok, turns out that in Ruby 2.0 refinements are "experimental" and are not fully supported.
What's our policy about supporting legacy versions?

mattbrictson:
I don't think we have an official policy, but I would feel a little uncomfortable making changes that require 2.1. Ruby 2.0 is not EOL yet.

kirs:
It's not end of life yet, but maintenance  and security releases will end
after Feb 24, 2016 for 2.0

robd:
I'm interested in our policy of which Ruby versions we support. Personally, I think we should support as many versions for as long as possible even after EOL has expired. I think we should only take the call to stop supporting an old version when there is some compelling language feature which we need to use. This is because I believe there are a lot of reasons why our users can get stuck on an old version of ruby and I think we should support them unless it makes our lives very difficult. In this case, since the monkey patched methods are only used in a limited number of places, I don't feel that refinements are the right way to go.

Personally, I don't feel strongly either way about whether we should remove the pollution of the global scope here. But assuming it is something we want to do, I'd like to discuss the usages of the 2 methods we're looking at - `symbolize_keys!` and `extract_options!`.

To remove the `symbolize_keys!` `Hash` patch, I wonder if we could add a private method `symbolize_keys(options)` to `Command` (since I believe this is the only place that uses that method). 

For the`extract_options!` method, it is used in the `Command` constructor and the `Abstract` backend. I believe the only usage of the `Command` constructor is [here](https://github.com/capistrano/sshkit/blob/8032db6a237b588df1c4c2df3e3a6a83c2befdd9/lib/sshkit/backends/abstract.rb#L149), and it seems like this would be cleaner if we removed the splat and passed the args and options separately:

```ruby
SSHKit::Command.new(args, options.merge({in: pwd_path, env: @env, host: @host, user: @user, group: @group}))
```

If we changed the constructor signature of `Command` then we would no longer need to `extract_options!` in the Command constructor. We could then add an `extract_options(args)` method on the `Abstract` backend.

The other thing is if we remove these methods from the global scope, any other gems / users who are relying on this method will no longer be able to use it. I don't know if this is an issue, but we should consider whether it needs to be a major release.

leehambley:
I tend to agree with Rob, as many for as long as reasonable, that said,
some are just too broken. I wish rubygems.org would report popularity
stats, and that CI tooling in general wouldn't be broken, so people
wouldn't have had such a negative reaction to the version tracking I tried
to bake in in 2015 (capistrano-stats gem). (CI tooling claiming to provide
a TTY caused a LOT of hung builds waiting for inout that never came on
stdin, checking for a TTY and dev device is the only sane way to make this
check, maybe I'll reintroduce it with a read timeout)

(sent from my phone, please excuse typos)

mattbrictson:
Is there a bug report tied to this change? If this is solving a purely hypothetical issue or is just for code cleanliness, then I don't think we should merge it right now.

kirs:
I agree with @robd. Let's work on a better way to replace core_ext.

Issue #326, opened 2016-02-02
Parallel executor performance is mostly linear


byroot:
We use capistrano to deploy on several hundred servers at once, and we noticed capistrano's performance was heavily tied to the number of servers.

To prove it I ran the following benchmark:

```ruby
require 'benchmark'
require 'csv'
task :bench do
  servers = roles(:borg).to_a
  timings = {}

  if ENV['PREWARM']
    on servers do
      execute :echo, 'pong > /dev/null'
    end
  end

  [1, 10, 50, 100].each do |count|
    timings[count] = Benchmark.measure do
      on servers.pop(count) do
        execute :echo, 'pong > /dev/null'
      end
    end
  end

  res = [%w(count user system real)]
  timings.each do |count, timing|
    res << [count, timing.utime, timing.stime, timing.real]
  end
  puts res.map(&:to_csv).join
end
```

Here are the results: [Spreadsheet](https://docs.google.com/spreadsheets/d/1iyJC3a2AAeqnWjH2uGKp2ZClEsVSxN2QB0D5v85LUBc/edit?usp=sharing)

<img width="903" alt="capture d ecran 2016-02-02 a 18 41 29" src="https://cloud.githubusercontent.com/assets/44640/12768297/c30d65b4-c9dc-11e5-9f25-65514aa0c001.png">

The first graph is with "cold" connections, meaning it's connection establishment plus the command execution. In the second graph, all the connections were pre established before the benchmark.

I'm still investigating to figure out where exactly the bottleneck (or bottlenecks) exactly is. I know the GIL is not for nothing, but capistrano / SSHKit being IO heavy, I think there is other reasons.

cc @kirs as well as @csfrancis

byroot:
Oh and I forgot, here's a profile of a similar benchmark with about 180 severs:

Cold: 
```
==================================
  Mode: wall(1000)
  Samples: 3592 (6.04% miss rate)
  GC: 390 (10.86%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       626  (17.4%)         620  (17.3%)     SSHKit::Runner::Parallel#execute
       511  (14.2%)         511  (14.2%)     block (3 levels) in Net::SSH::KnownHosts#keys_for
       884  (24.6%)         306   (8.5%)     block (2 levels) in Net::SSH::KnownHosts#keys_for
[...]

PR #327, opened 2016-02-03
Allow to disable both connection pooling and pruning idependently

byroot:
Ref: https://github.com/capistrano/sshkit/issues/326

By profiling capistrano, it occurred that connection pruning was a significant performance hotspot.

See https://github.com/capistrano/sshkit/issues/326#issuecomment-178939657
> After investigation it's clear that the ConnectionPool put way to much zeal into pruning expired connection.
> The prune_expired method that iterates over all the connections is called both on checkin and on checkout, so it's executed twice per host and per command.
> e.g. if you execute 5 commands, on 10 servers, it will be executed 2 * 5 * 10, 100 times! Also note that it synchronize a mutex, so it's very probable that it kills limits the concurrency.

So much that 12% of the execution time (on a micro benchmark but still) was spent checking expirations.

I think that for many cases just checking is the connection is closed on checkout is enough.

I ran the same benchmark a saw some interesting performance improvements (even though it's always hard to be put an exact number because of the network).
capture d ecran 2016-02-03 a 00 12 37

NB: This is somewhat backwards incompatible. `idle_timeout = 0` used to disable connection pooling entirely by always reaping connections, while now it actually let connections live indefinitely until they are closed.

If a version bump is not an option, I can make `idle_timeout = 0` disable pooling as well, but then we need something like `idle_timeout = false` to disable connection pruning.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,8 @@ This file is written in reverse chronological order, newer releases will
 appear at the top.
 
 ## `master` (Unreleased)
-
+  * `SSHKit::Backend::Netssh.pool.idle_timeout = 0` doesn't disable connection pooling anymore,
+    only connection expiration. To disable connection polling use `SSHKit::Backend::Netssh.pool.enabled = false`
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
   * make sure working directory for commands is properly cleared after `within` blocks
--- a/README.md
+++ b/README.md
@@ -500,11 +500,16 @@ seconds. This timeout can be changed as follows:
 SSHKit::Backend::Netssh.pool.idle_timeout = 60 # seconds
 ```
 
-If you suspect the connection pooling is causing problems, you can disable the
-pooling behaviour entirely by setting the idle_timeout to zero:
+If you suspect the connection pruning is causing problems, you can disable the
+expiration behaviour entirely by setting the idle_timeout to zero:
 
 ```ruby
 SSHKit::Backend::Netssh.pool.idle_timeout = 0 # disabled
+```
+
+Or even disable the pooling entirely, but it will severely reduce performance:
+```
+SSHKit::Backend::Netssh.pool.enabled = false # disabled
 ```
 
 ## Tunneling and other related SSH themes
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -19,28 +19,35 @@ module SSHKit
 
     class ConnectionPool
 
-      attr_accessor :idle_timeout
+      attr_accessor :idle_timeout, :enabled
 
       def initialize
         self.idle_timeout = 30
+        self.enabled = true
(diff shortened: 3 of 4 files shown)


byroot:
WALL profile of existing connections being used before the patch:
```
==================================
  Mode: wall(1000)
  Samples: 3860 (1.40% miss rate)
  GC: 442 (11.45%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       567  (14.7%)         567  (14.7%)     SSHKit::Backend::ConnectionPool::Entry#expired?
       347   (9.0%)         347   (9.0%)     Net::SSH::Compat.io_select
       341   (8.8%)         341   (8.8%)     block (2 levels) in Net::SSH::Connection::Channel#forward_local_env
       681  (17.6%)         106   (2.7%)     block (2 levels) in SSHKit::Backend::ConnectionPool#prune_expired
        90   (2.3%)          90   (2.3%)     Net::SSH::Transport::State#update_next_iv
        89   (2.3%)          89   (2.3%)     Net::SSH::Transport::HMAC::Abstract.digest_class
```

After the patch: 
```
==================================
  Mode: wall(1000)
  Samples: 3033 (2.38% miss rate)
  GC: 359 (11.84%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
       347  (11.4%)         347  (11.4%)     Net::SSH::Compat.io_select
       345  (11.4%)         345  (11.4%)     block (2 levels) in Net::SSH::Connection::Channel#forward_local_env
        98   (3.2%)          91   (3.0%)     SSHKit::Runner::Parallel#execute
        87   (2.9%)          87   (2.9%)     Net::SSH::Transport::HMAC::Abstract.mac_length
        86   (2.8%)          86   (2.8%)     Net::SSH::Transport::HMAC::Abstract.digest_class
        91   (3.0%)          75   (2.5%)     Net::SSH::Buffer#read
       119   (3.9%)          72   (2.4%)     Net::SSH::BufferedIo#fill
        70   (2.3%)          70   (2.3%)     Net::SSH::Transport::State#update_next_iv
       116   (3.8%)          62   (2.0%)     Net::SSH::BufferedIo#send_pending
        56   (1.8%)          56   (1.8%)     block in Net::SSH::Transport::PacketStream#enqueue_packet
```

So as expected the next bottleneck is `IO.select`, but it's already a big win.

byroot:
I pushed a new commit using my SSHKit fix. Please re-review.

kirs:
Looks good for me.

@leehambley @mattbrictson

mattbrictson:
This looks good to me as well.

@byroot Can you add a CHANGELOG entry explaining the "somewhat backwards incompatible" behavior so that users are not caught off guard?

Once that's in I would like to merge this.

Thanks!

byroot:
Sure!

byroot:
@mattbrictson done!

mattbrictson:
:tada:

PR #335, opened 2016-02-29
Make pwd_path public

drewbug:


mattbrictson:
Can you provide some more background on why `pwd_path` should be public?

drewbug:
Keeping track of it myself has been a pain, especially when trying to write utility classes that treat the backend as a kindof remote Kernel class. I don't understand why it was made private to begin with.

leehambley:
iirc it was done this way because it was literally only ever needed as part of that method. If there's a case to be made, and we can test that it works extracted into it's own method, I'd be fine with the change.

However @drewbug we need you to look at [this document](https://github.com/capistrano/capistrano/blob/master/DEVELOPMENT.md) to get code merged.

leehambley:
bump @drewbug 

> However @drewbug we need you to look at this document to get code merged.

PR #330, opened 2016-02-11
Provide an alternative implementation of Net::SSH::KnownHost in ssh_options

byroot:
As discussed in https://github.com/capistrano/sshkit/issues/326#issuecomment-179284023
Net::SSH re-parse the known_hosts files every time it needs to lookup for a known key.
This alternative implementation parse it once and for all, and cache the result.

NB: This is not complete yet, I need to also make `test_known_hosts_for_when_an_host_hash_is_recognized` pass.

cc @mattbrictson @TiagoCardoso1983

PR #339, opened 2016-03-11
Cache known hosts lookup by default

byroot:
Followup to https://github.com/capistrano/sshkit/pull/330

To be merged at some point.

cc @mattbrictson

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * Known hosts lookup optimization is now enabled by default. @byroot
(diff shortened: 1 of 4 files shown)


leehambley:
TIL about http://ruby-doc.org/stdlib-2.3.0/libdoc/mutex_m/rdoc/Mutex_m.html :) thanks :+1:

mattbrictson:
@byroot No hurry, but this will also need to be rebased at some point so we can merge it for a future release.

byroot:
Done!

mattbrictson:
Thanks. Could you add a brief CHANGELOG entry?

byroot:
Done as well!

PR #343, opened 2016-04-22
Explicitly close the connection pool using at_exit

### LINKED ISSUE #342: zlib(finalizer): the stream was freed prematurely. when using compression
```ruby
require 'sshkit'
require 'sshkit/dsl'

SSHKit::Backend::Netssh.config.ssh_options = {compression: true, compression_level: 9}

on 'localhost' do
  ls_output = capture(:ls, '-l')
end
```

It outputs:
```
...
zlib(finalizer): the stream was freed prematurely.
```

the issue is that sshkit doesn't close the net-ssh connection which then cannot close the zlib stream which has finalizer and complains.

as a workaround
```sh
SSHKit::Backend::Netssh.pool.close_connections
```
can be added thanks to https://github.com/capistrano/sshkit/pull/285
see also https://github.com/net-ssh/net-ssh/issues/284

mattbrictson:
This fixes the `zlib(finalizer): the stream was freed prematurely` warning that is printed by net-ssh when Ruby exits without closing the SSH connection first.

Fixes #342.

I could have placed the `at_exit` registration inside `ConnectionPool#initialize` (as I [originally suggested](https://github.com/capistrano/sshkit/issues/342#issuecomment-210202285)), but then that would mean all instances of ConnectionPool would never be garbage collected (the `at_exit` block would retain a reference to `self`). Doing it in the `Netssh` class definition seemed like a better place to put it.

/cc @leehambley @mfazekas

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * SSHKit now explicitly closes its pooled SSH connections when Ruby exits;
+    this fixes `zlib(finalizer): the stream was freed prematurely` warnings
+    [PR #343](https://github.com/capistrano/sshkit/pull/343) @mattbrictson
   * Allow command map entries (`SSHKit::CommandMap#[]`) to be Procs
     [PR #310]((https://github.com/capistrano/sshkit/pull/310)
     @mikz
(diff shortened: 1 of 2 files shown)


mfazekas:
looks great, thank!

leehambley:
> I could have placed the at_exit registration inside ConnectionPool#initialize (as I originally suggested), but then that would mean all instances of ConnectionPool would never be garbage collected (the at_exit block would retain a reference to self). 

Hrm, brb - fixing a bunch of projects where I did this wrong! Thanks Matt. 

I'm not the biggest fan of `require 'English'` especially in library code, but fair enough!

Issue #43, opened 2013-10-31
Nesting 'with' fails


clintonb:
I have some example code below from my Capistrano v3 tasks for Django. It seems that nesting 'with' calls is broken. Is this functionality supposed to be supported? The docs state "One will notice that it's quite low level, but exposes a convenient API, the as()/within()/with() are nestable in any order, repeatable, and stackable."

My current work around is to wrap the body of with_virtualenv in a begin-rescue block that captures the NameError and check for the message "instance variable @_env not defined".

Note that my intention is to keep the Django and virtualenv tasks in separate files/gems, hence my not simply merging the with calls.

```
def with_virtualenv(&block)
  with path: "#{fetch(:virtualenv_dir)}/bin:$PATH" do
    block.call
  end
end

def django_manage(*args)
  within release_path do
    with_virtualenv do
      with app_environment: fetch(:stage) do
        execute :python, 'manage.py', args
      end
    end
  end
end

# Results in "instance variable @_env not defined"
namespace :django do
  task :migrate do
    on primary fetch(:migration_role) do
      django_manage 'syncdb', '--noinput', '--migrate'
    end
  end
end
```

DylanFrese:
Should be fixed with PR #345

leehambley:
If someone can confirm, I'd be glad to close this. I'll swing by again in a few days and check.

ArturT:
@leehambley  I had a problem during deploy with capistrano and whenever:update_crontab.

```
cap aborted!
NameError: instance variable @_env not defined

Tasks: TOP => whenever:update_crontab
```

I switched to sshkit master branch and the PR https://github.com/capistrano/sshkit/pull/345 that was merged solved my problem.

ansonhoyt:
@leehambley I echo @ArturT, updating to master fixed it for me.

If it helps confirm, here's output with `--trace` before I switched to master:
```
ruby 2.3.0p0 (2015-12-25 revision 53290) [x86_64-linux]
** Invoke bundler:map_bins (first_time)
** Invoke passenger:bundler:hook (first_time)
** Execute passenger:bundler:hook
** Execute bundler:map_bins
** Invoke deploy:set_rails_env (first_time)
** Execute deploy:set_rails_env
** Invoke deploy:set_linked_dirs (first_time)
** Execute deploy:set_linked_dirs
** Invoke deploy:set_rails_env
** Invoke whenever:update_crontab (first_time)
** Execute whenever:update_crontab
[...]

Issue #176, opened 2014-09-04
instance variable @_env not defined error


tamadamas:
```
SSHKit::Runner::ExecuteError: Exception while executing on host: instance variable @_env not defined
~/.rvm/gems/ruby-1.9.3-p547@ges/gems/sshkit-1.5.1/lib/sshkit/backends/abstract.rb:88:in `remove_instance_variable'
~/.rvm/gems/ruby-1.9.3-p547@ges/gems/sshkit-1.5.1/lib/sshkit/backends/abstract.rb:88:in `with'
```

And it falls on next task

```
desc 'Create backup of the DB'
  task :backup do
    on roles(:db) do
      within release_path do
        with rails_env: fetch(:rails_env) do
          with 'BACKUP_PATH' => shared_path.join('backups') do
            execute :rake, 'db:backup'
          end
        end
      end
    end
  end
```

And task completes successfully and I can see correct dump.

PR #345, opened 2016-04-29
Change @_env to local variable

DylanFrese:
In the SSHKit Abstract backend, the 'with' method declared an instance variable (@_env) that was only used in that method, and ensured it was unset before returning. This would cause errors when nesting 'with' calls, as the outer call would try to use and then unset @_env after the inner call already unset it. It has been changed to a regular local variable, as the desired behaviour is identical to using a local variable anyway.

This fixes Issue #43 and Issue #176

### DIFF
--- a/lib/sshkit/backends/abstract.rb
+++ b/lib/sshkit/backends/abstract.rb
@@ -83,12 +83,11 @@ module SSHKit
       end
 
       def with(environment, &_block)
-        @_env = (@env ||= {})
-        @env = @_env.merge environment
+        _env = (@env ||= {})
+        @env = _env.merge environment
         yield
       ensure
-        @env = @_env
-        remove_instance_variable(:@_env)
+        @env = _env
       end
 
       def as(who, &_block)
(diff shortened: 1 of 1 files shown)


leehambley:
Could you check the build please @DylanFrese ?

DylanFrese:
@leehambley It looks like RuboCop is complaining about the fact that '_env' is a used variable and starts with an underscore. I did this because that's what the original variable was named (@_env), but I'll change it.

DylanFrese:
@leehambley Done

leehambley:
Thanks, LGTM I just would ask you to amend the commit again and credit yourself in the CHANGELOG if you're interested in being credited, if not just let me know and I'll merge the PR as-is.

DylanFrese:
You're free to merge it as-is.

gucki:
@leehambley Could you please release a version with this fixed? For example capistrano whenever fails due to this bug.

oprema:
Yes this would be wonderful if @leehambley could fix this. My capistrano fails with the same error. Thanks.

mattbrictson:
I plan on releasing 1.11.0 in a few hours, which will include this fix.

PR #346, opened 2016-05-13
One of the examples in the README doesn't work

ttanimichi:


mattbrictson:
Nice catch. Thanks! ✨

PR #347, opened 2016-05-17
Support future ruby

ksss:
On ruby-head, Fixnum and Bignum are Unified.

https://github.com/ruby/ruby/commit/f9727c12cc8fbc5f752f5983be1f14bb976e5a13

And capistrano fail with this error when use ruby-head 

```
NoMethodError: undefined method `>=' for nil:NilClass
```

This patch will support both old and new ruby.

### DIFF
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -131,9 +131,9 @@ module SSHKit
 
     def verbosity
       if (vb = options[:verbosity])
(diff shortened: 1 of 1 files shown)


mattbrictson:
Interesting! Thanks for the heads-up and the contribution. ✨

leehambley:
Yep, great catch, thanks a lot!

PR #349, opened 2016-05-20
Fix colorized log alignment.

xavierholt:
The original intent was pretty obviously to have the log level right-justified in a six-width column - this works great in the non-colorized tests - but the non-printing characters for colorization are counted as part of the level string's length, breaking the alignment.  Padding before colorizing dodges the problem.

Let me know if there's anything else I can / should do!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ appear at the top.
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
   * Known hosts lookup optimization is now enabled by default. @byroot
+  * Fixed colorized output alignment in Logger::Pretty. @xavierholt
+    [PR #349](https://github.com/capistrano/sshkit/pull/349)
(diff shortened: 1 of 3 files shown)


leehambley:
Credit yourself in the changelog and squash the commit if you like, if not, I'd be happy to merge as-is!

xavierholt:
Updated!  May have messed up the squash (sorry - never done that particular bit of git witchery before), but go ahead and merge any time you like.  Cheers!

mattbrictson:
@xavierholt Looks like your squash didn't work; what we're looking for is all of your changes in a single commit (instead of 3).

However! GitHub has a shiny new "merge and squash" button available to repository maintainers that is supposed to do the job for us, and I will give that a shot now.

Thanks for the fix!

mattbrictson:
"Merge and squash" worked as advertised. 👍

Issue #344, opened 2016-04-22
Running SSHKit tests with Rake 11.0 reveals many Ruby warnings


mattbrictson:
Rake 11.0 turns on Ruby warnings by default. With warnings enabled, running SSHKit's test suite results in a flood of messages like this:

```
lib/sshkit/runners/group.rb:15: warning: instance variable @group_size not initialized
lib/sshkit/host.rb:59: warning: instance variable @local not initialized
lib/sshkit/backends/abstract.rb:141: warning: instance variable @pwd not initialized
lib/sshkit/backends/abstract.rb:149: warning: instance variable @env not initialized
lib/sshkit/backends/abstract.rb:149: warning: instance variable @user not initialized
lib/sshkit/backends/abstract.rb:149: warning: instance variable @group not initialized
```

We could:

* Treat these warnings as legitimate and try to fix them; or
* Turn off Ruby warnings

I haven't had time to think this through yet, so for now I'm just opening it for discussion.

Your thoughts?

leehambley:
Warnings are legit, we should solve them, but if it's only a test problem, I wouldn't let it block you from 1.10 release?

mattbrictson:
Nope, this is definitely not a blocker. Just wanted to make note of it before I forgot.

PR #350, opened 2016-05-21
Fix uninitialized instance variable / overridden function warnings.

xavierholt:
Found myself cleaning up some Ruby warnings while testing a branch earlier today, so I thought I'd finish the job and help out with https://github.com/capistrano/sshkit/issues/344

Couple notes:
 - Most changes are simply setting things to `nil` in `initialize` methods.
 - Identical `unindent` functionality is provided by the unindent gem, so the monkeypatch version isn't needed any more.
 - Part of my fix makes `group_size` and `wait_interval` publicly readable on the runners.  It seemed odd to have them writeable but not readable, so I felt okay making the visibility change.  If that's not okay, let me know and I'll revert that part.


leehambley:
LGTM :) thanks for the contribution - Should I merge right now, or would you like to credit yourself in the changelog ?

robd:
Hey - apologies for chiming in out of nowhere! 

One thing I noticed here (unless I missed something) is that the `unindent` method is only called in one place (`test/functional/backends/test_netssh.rb`). I wondered if it would be better to remove the SSHKit dependency on the `unindent` Gem, and move the unindent method from `test/unit/core_ext/test_string.rb` to `test_netssh.rb`.

I guess the only other thing to check is whether the build is passing - see discussion here: https://github.com/capistrano/sshkit/issues/351

xavierholt:
@robd:  Updated with an even further simplification - scrap the unindent method entirely!

@leehambley:  No need - I got credit for my actual bugfix, and this is just trivial cleanup.  I'm okay to merge whenever.  Not sure if https://github.com/capistrano/sshkit/issues/351 is relevant.  It'll pass CI; the issue only shows up when running on a host with a non-empty known_hosts file...  Just let me know if you'd like me to rebase after the fix is in.

robd:
> scrap the unindent method entirely!

👍 Even better!

mattbrictson:
#351 is now fixed in master. Can you rebase, please?

xavierholt:
@mattbrictson:  Rebased.  Cheers!

mattbrictson:
Acceptance tests pass. ✅ 

@leehambley I'll leave the merge to you.

leehambley:
Thanks @xavierholt and for the sanity checking @mattbrictson

Issue #351, opened 2016-05-21
test_ssh_option_merge is failing on master (Vagrant functional test)


mattbrictson:
Getting this error when running the full test suite:

```
$ vagrant up
$ bundle _1.10.5_ exec rake
[snip]
  test_ssh_option_merge                                           FAIL (0.05s)
Minitest::Assertion:         --- expected
        +++ actual
        @@ -1 +1 @@
        -{:forward_agent=>false, :paranoid=>true}
        +{:known_hosts=>#<SSHKit::Backend::Netssh::KnownHosts:0xXXXXXX @_mutex=#<Thread::Mutex:0xXXXXXX>, @files={"~/.ssh/known_hosts"=>#<SSHKit::Backend::Netssh::KnownHostsKeys:0xXXXXXX
[snip]
```

Looks like the break is due to the merge of #339.

@byroot Since that was your commit, can you take the first shot at fixing this?

The test in question is `SSHKit::Backend::TestNetssh#test_ssh_option_merge`.

PR #353, opened 2016-05-22
Fix default options test 

byroot:
Fix: https://github.com/capistrano/sshkit/issues/351

cc @mattbrictson


mattbrictson:
Thanks! 🙇

PR #356, opened 2016-06-17
Fix interaction handler docs in the README

### LINKED ISSUE #355: Can't use MappingInteractionHandler
Stats:
  * SSHKit 1.11.0
  * Capistrano 3.4.0
  * Ruby 2.3.0

I need to run a remote mysql dump command without sending the password, something like:

```bash
ssh user@server "/usr/bin/mysqldump -p ..."
```

Which when run manually in the command-line works as expected, ie. it asks:

    Enter password:

So I created a very simple Capistrano task to try to use the MappingInteractionHandler as explained in the [README](https://github.com/capistrano/sshkit#exploratory-logging) like:

```ruby
task :some_task do
    begin
        on roles(:app)[0] do
            execute("/usr/bin/mysqldump -p ...", SSHKit::MappingInteractionHandler.new({}, :debug))
        end
    end
end
```

But the task simply hangs after the INFO line with "Running ...". 

I also tried with the expected command output:

```ruby
execute("/usr/bin/mysqldump -p ...", SSHKit::MappingInteractionHandler.new({"Enter password: " => "pw"}, :debug))
```

With the same result. Am I missing something?

I also noticed that no debug log seems to be shown on screen (is it actually supposed to be shown on the screen or to go to some log file?).

robd:
Closes #355

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ desirable.
 first argument before attempting to find it in the *command map*.
 
 ## Interactive commands
-> (BETA) (Added in version #.##)
+> (Added in version 1.8.0)
 
 By default, commands against remote servers are run in a *non-login, non-interactive* ssh session.
 This is by design, to try and isolate the environment and make sure that things work as expected,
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks for taking care of this! 🙌

robd:
No probs - sorry I haven't been involved recently - super hectic this end.

Issue #357, opened 2016-06-17
ArgumentError: invalid option(s): known_hosts with net-ssh 3.0.2


mattbrictson:
I saw this reported on Stack Overflow, and it seems like a legitimate bug introduced in sshkit 1.11.0.

https://stackoverflow.com/questions/37865299/capistrano-error-invalid-optionss-known-hosts

To summarize: starting with sshkit 1.11.0, we now pass a `:known_hosts` option to net-ssh by default (see #339). Unfortunately it seems that not all versions of net-ssh allow this. And instead of ignoring the option, net-ssh raises this error, at least in net-ssh 3.0.2:

`ArgumentError: invalid option(s): known_hosts`

The solution seems to be to upgrade to net-ssh 3.1.1.

@byroot Does this bug report seem legitimate? What can we do to fix this issue (other than change our gemspec to require newer net-ssh, which seems drastic)?

casperisfine:
Hum, good point. I think we could check `Net::SSH::VALID_OPTIONS`, I'll PR something.

PR #358, opened 2016-06-17
Add check for :known_hosts support in net-ssh

casperisfine:
Should fix https://github.com/capistrano/sshkit/issues/357

Disclaimer: I haven't had the time to test it properly yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * Fixed compatibility with net-ssh > 3.1.0. @byroot
+    [#357](https://github.com/capistrano/sshkit/issues/357)
 
 ## [1.11.0][] (2016-06-14)
 
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -33,8 +33,14 @@ module SSHKit
 
         private
 
-        def default_options
-          @default_options ||= {known_hosts: SSHKit::Backend::Netssh::KnownHosts.new}
+        if Net::SSH::VALID_OPTIONS.include?(:known_hosts)
+          def default_options
+            @default_options ||= {known_hosts: SSHKit::Backend::Netssh::KnownHosts.new}
+          end
+        else
+          def default_options
+            @default_options ||= {}
+          end
         end
       end
 


mattbrictson:
Given that CI passes, this code should be fine with the latest net-ssh. Can someone manually test that it does the right thing for 3.0.2 (which doesn't support `:known_hosts`)?

mattbrictson:
Also: a CHANGELOG entry would be much appreciated!

casperisfine:
> Also: a CHANGELOG entry would be much appreciated!

Snap, I always forgot those. (by the way I'm @byroot)

casperisfine:
Ok, just tested it indeed solves the problem: https://gist.github.com/casperisfine/6810e2f1df54139f9c3f0cab84876b00

```bash
byroot@haunted-house:tmp:$ ruby ssh.rb 
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Resolving dependencies...
Using net-ssh 3.0.2
Using bundler 1.12.3
Using net-scp 1.2.1
Using sshkit 1.11.0
  INFO [f61cef01] Running echo Hello on localhost
/Users/byroot/.gem/ruby/2.1.8/gems/net-ssh-3.0.2/lib/net/ssh.rb:197:in `start': invalid option(s): known_hosts (ArgumentError)
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/connection_pool.rb:59:in `call'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/connection_pool.rb:59:in `with'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/netssh.rb:149:in `with_ssh'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/netssh.rb:102:in `execute_command'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/abstract.rb:141:in `block in create_command_and_execute'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/abstract.rb:141:in `tap'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/abstract.rb:141:in `create_command_and_execute'
	from /Users/byroot/.gem/ruby/2.1.8/gems/sshkit-1.11.0/lib/sshkit/backends/abstract.rb:74:in `execute'
	from ssh.rb:22:in `<main>'
byroot@haunted-house:tmp:$ FIX=1 ruby ssh.rb 
Fetching git://github.com/shopify/sshkit.git
Fetching gem metadata from https://rubygems.org/
Fetching version metadata from https://rubygems.org/
Resolving dependencies...
Using net-ssh 3.0.2
Using bundler 1.12.3
Using net-scp 1.2.1
Using sshkit 1.11.0 from git://github.com/shopify/sshkit.git (at handle-older-net-ssh@fd39260)
  INFO [5ffa3ef7] Running echo Hello on localhost
  INFO [5ffa3ef7] Finished in 0.255 seconds with exit status 0 (successful).
```

mattbrictson:
Excellent work. Thanks! 🙇

Issue #359, opened 2016-07-15
`any?` wrong number of arguments upon upgrading 1.10.0 -> 1.11.1 


PatrickatPaperlessPCS:
I have a rails app that I've used Capistrano to deploy with several times.  I did a fairly major update on the app that included an update of Capistrano-rails from 1.1.6 to 1.1.7 and SSHkit from 1.10.0 to 1.11.1.  Now my `cap production deploy` fails with what seems like an sshkit issue.  

Here's my error output:

```
** Invoke deploy:check (first_time)
** Execute deploy:check
** Invoke git:check (first_time)
** Invoke git:wrapper (first_time)
** Execute git:wrapper
00:00 git:wrapper
      01 mkdir -p /tmp/dentalimager/
cap aborted!
ArgumentError: wrong number of arguments (0 for 1)
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/capistrano-3.5.0/lib/capistrano/dsl/env.rb:16:in `any?'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/sshkit-1.11.1/lib/sshkit/host.rb:84:in `block in netssh_options'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/sshkit-1.11.1/lib/sshkit/host.rb:83:in `tap'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/sshkit-1.11.1/lib/sshkit/host.rb:83:in `netssh_options'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/capistrano-3.5.0/lib/capistrano/configuration/server.rb:59:in `netssh_options'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/sshkit-1.11.1/lib/sshkit/backends/netssh.rb:159:in `with_ssh'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/sshkit-1.11.1/lib/sshkit/backends/netssh.rb:108:in `execute_command'
/home/patrick/.rbenv/versions/2.2.1/lib/ruby/gems/2.2.0/gems/sshkit-1.11.1/lib/sshkit/backends/abstract.rb:141:in `block in create_command_and_execute'
[...]

PR #360, opened 2016-07-18
Fix for #359

xavierholt:
Make sure that `Host#keys` always returns an array, avoiding https://github.com/capistrano/sshkit/issues/359

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * Fixed a crash occurring when `Host@keys` was set to a non-Enumerable.
+    @xavierholt [PR #360](https://github.com/capistrano/sshkit/pull/360)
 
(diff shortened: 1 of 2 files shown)


mattbrictson:
Thanks for the quick turnaround! Can you also add a changelog entry? Then I'll merge.

xavierholt:
No problem!  Will do shortly.  But while I'm making another edit, do you think it should use the `Array(x)` wrapper in `#keys=` instead of `#keys`?  It would only make a difference in edge cases like `host.keys << 'my key'` - is there an expected behaviour?

mattbrictson:
Honestly I'm not sure what all the real-world use cases are for `Host#keys`, so I think it is safest to simply restore how it worked before, i.e. with the wrapper in `#keys`.

xavierholt:
There you go!  Changelog in, everything else left as-is (as-was?).

mattbrictson:
Thanks! 🙇

leehambley:
Thanks for the fixes and quick merge both of you :)

Issue #362, opened 2016-07-21
Upgrading 1.10.0 to 1.11.1 breaks connecting to new hosts.


xavierholt:
Upgraded SSHKit this morning and found that I could no longer connect to hosts that were not in my `~/.ssh/known_hosts` file.  I could downgrade and connect then upgrade again, and connecting worked just fine.

```
[kevin.burk@KBURK-02 argus]$ bundle update | grep sshkit
Using sshkit 1.11.1
[kevin.burk@KBURK-02 argus]$ bundle exec rake k8sdev uruk run cmd="uptime"
INFO   [28c16682] Running /usr/bin/env uptime on 10.52.186.21
DEBUG  [28c16682] Command: /usr/bin/env uptime
INFO   [e4ed48aa] Running /usr/bin/env uptime on 10.52.186.22
DEBUG  [e4ed48aa] Command: /usr/bin/env uptime
INFO   [a8d5f190] Running /usr/bin/env uptime on 10.52.186.23
DEBUG  [a8d5f190] Command: /usr/bin/env uptime
rake aborted!
Net::SSH::HostKeyMismatch: fingerprint 41:31:33:63:66:9d:34:a9:de:35:f2:4d:03:ee:4d:b4 does not match for "10.52.186.21,127.0.0.1"
/Users/kevin.burk/Code/argus/lib/tasks/helpers.rb:40:in `block (3 levels) in <top (required)>'
Tasks: TOP => run
(See full trace by running task with --trace)
[kevin.burk@KBURK-02 argus]$ vim Gemfile
[kevin.burk@KBURK-02 argus]$ bundle update | grep sshkit
Using sshkit 1.10.0 (was 1.11.1)
[kevin.burk@KBURK-02 argus]$ bundle exec rake k8sdev uruk run cmd="uptime"
INFO   [7178e030] Running /usr/bin/env uptime on 10.52.186.23
DEBUG  [7178e030] Command: /usr/bin/env uptime
INFO   [5dfe4842] Running /usr/bin/env uptime on 10.52.186.22
DEBUG  [5dfe4842] Command: /usr/bin/env uptime
INFO   [c8d074b0] Running /usr/bin/env uptime on 10.52.186.21
DEBUG  [c8d074b0] Command: /usr/bin/env uptime
DEBUG  [7178e030] 	 19:57:56 up  1:37,  0 users,  load average: 0.02, 0.05, 0.05
[...]

PR #364, opened 2016-07-27
Fix KnownHostsKeys#keys_for to match on the entire hostlist

byroot:
@xavierholt I believe you were spot on about https://github.com/capistrano/sshkit/issues/362#issuecomment-234742497. Woudl you mind testing this branch with your setup. I think that should fix your issue.


mattbrictson:
Thanks for the PR! It looks like @xavierholt has confirmed this fixes #362, so I would like to merge this in.

@byroot Can you add a changelog entry?

byroot:
Done!

mattbrictson:
Thank you! 🙇

PR #365, opened 2016-07-28
Updated README on how as() works

Dbz:
Hello! This is a very minor pull request to fix the readme on how `as()` works. It caused us some confusion, so I figured this might help someone else.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,16 +94,16 @@ When used inside a block in this way, `as()` and `within()` will guard
 the block they are given with a check.
 
 In the case of `within()`, an error-raising check will be made that the directory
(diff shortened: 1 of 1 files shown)


mattbrictson:
Thanks for the PR! Good catch and I agree this doc needs to be changed.

However when I look at the current [source code](https://github.com/capistrano/sshkit/blob/df8ac2e41577a52806d6708701ff8d2fbddcc737/lib/sshkit/command.rb#L171), it seems that the actual `as` effect is more like this:

```
sudo -u <user> -- sh -c 'whoami'
```

Dbz:
Ah you are correct. I was only looking in `abstract.rb` because I'm not very familiar with the codebase.

mattbrictson:
Thanks! 🙌

PR #368, opened 2016-09-02
Proposal: Docker backend

sugi:
Hello,

I'd like to propose new backend for docker to sshkit.

This backend add a direct connectivity to docker container  (through docker socket without ssh connection). User can deploy their app into docker container by capistrano with this backend.

How do you feel about this?

Example is here; https://github.com/sugi/capistrano-deploy_into_docker


leehambley:
We are currently deprecating the `local` (`run_locally`) backend, because it operates semantically differently and we don't find it suitable for it's intended purpose.

Much of the motivation is that the semantics are simply different, environments don't work the same way, synchronisation is difficult, etc.

My personal opinion is that a customised Docker backend that specifically ignores other container types and does not use SSH has no place in SSHKit, but perhaps some of the other maintainers can share their opinion.

You could almost certainly maintain this as a separate Gem looking at the implementation. Given the apparent quality of the implementation it is surely useful, but I don't think it has a place being in the core of something called *SSHKit*.

@mattbrictson @will-in-wi ?

will-in-wi:
I have a couple of thoughts…

1. This is a quite impressive contribution. Whatever we do, I want to try and make sure this wasn't a waste of time for @sugi. 👍 

2. I don't know much about Docker. I've messed with it, and it seems like the "right" way of deploying apps via Docker is to have a Dockerfile in the application repo which defines the tasks that Capistrano is presently doing. I'm curious (although this isn't as relevant to how this PR is handled) why Capistrano is more useful than the alternative method.

3. I do see this as sparking a question of a future path for SSHKit. I agree that Docker support seems out of scope for SSHKit. One option is to rename SSHKit to DeployKit or some such, and incorporate things like this into it. Another option would be to define an API, either in Capistrano or SSHKit which allows third-party plugin backends. To some extent heretofore, the "Capistrano API" has been a mix of DSLs from Capistrano and SSHKit. If we wanted to extract this in some way, we ought to try and minimize the "leaky" use of the SSHKit library by Capistrano end users, and probably wrap it (DeployKit?) in something that can have multiple backends.

I'm not sure what to do with this. The contribution is quite nice, and I hate to simply stamp it with "out-of-scope" (I've had that done to me before as a community contributor and it is demotivating). But as is, it kind of is. I'm trying to figure out what the best approach for making contributions like this more possible in future without forking. I'd be more than happy to work on this.

My 2¢.

leehambley:
Thanks @will-in-wi totally echoing the way I feel, I'd gladly have stomped on the idea as "out of scope" if it weren't for the quality of the implementation. 

I want to see this be useful... I'm not sure if we should reposition SSHKit, a lot of the concepts we have are pretty specific to SSHKit, and we *kind of* already had that discussion deciding if `run_locally` was a good idea to keep, deprecate or improve..

sugi:
Thank you for your comments! 

Yeah, I agree with "out of scope", too...

I use SSHKit as capistrano backend.
When I tried to add docker support, I think it's better that I implement the feature as a SSHKit backend.
However, "Docker backend" is weird as "SSHKit" actually.

On the other hand, SSHKit is easy to use and has very good DSL to execute commands transparently via ssh,  ...and other transports (includes local), maybe.
And capistrano is still convenient to deploy rails app to local or docker.
I'd like to use capistrano to deploy rails app handy.

Anyway, I don't rush to merge the feature and it's OK if not.

mattbrictson:
@leehambley @will-in-wi I will defer to you guys on this one. My knowledge of Docker is very limited. If this were to be merged, it is not code that I could maintain on my own.

Since `SSHKit.config.backend` is already pluggable, couldn't the Docker backend be distributed as its own gem? The only changes of core SSHKit classes in this PR are (1) syntactic sugar for `host.rb` and (2) auto-switching backends based on that sugar. Neither is required for this to work, so it seems like the Docker-specific code is already cleanly separated.

will-in-wi:
@sugi: Do you think that this is something you could cleanly distribute as its own pluggable Gem? 

Are there changes which could be made to SSHKit that would make it easier for you to integrate cleanly?

sugi:
Oh, I see :satisfied:. I missed `SSHKit.config.backend`!

It's good  that I separate gem and user switch backend by the config variable.
Then, I withdraw this PR.

Thank you all!  :clap:

will-in-wi:
@sugi Thanks for your work on this! When you do complete a pluggable gem, I'd appreciate it if you would link to it here, just so that anyone else coming across this can see what the result was.

sugi:
Hello,

I published 2 gems. I'm happy if anyone is interested. 

 * [sshkit-backend-docker](https://github.com/sugi/sshkit-backend-docker)  - Docker backend for SSHKit that I sent proposal on this ticket.
 * [capistrano-deploy_into_docker](https://github.com/sugi/capistrano-deploy_into_docker) - Support tasks to use SSHKit Docker backend with capistrano.

Finally, thank you again for awesome SSHKit!

Startouf:
Hello, I'd just like to contribute with my view of docker with capistrano/SSH kit (Or deploy kit ;-) ).

One advantage Capistrano has over a Dockerfile, as @sugi highlighted, is its awesome synergy with rails, and managing different deployment environments is much easier (unless someone wants to keep one dockerfile per environment, which can be tricky hummmmm). Docker was made to abstract system/platform dependencies, not application dependencies

My deployment flow seems pretty legit (?)

1. Building a suitable base image with a generic Dockerfile and/or reuse one of the available images out there like `passenger-docker`. This should be done without `cap`, and once for good, or unless we need to add new system dependencies in which case we need to rebuild an image. The application should NOT be added to the container at this point

2. Start a container from this image. And deploy into it using @sugi 's awesome [deploy_into_docker](https://github.com/sugi/capistrano-deploy_into_docker) backend. We can make new images with new version numbers. However we can keep deploying each time to the (~same) container : it's like we're deploying new revisions, but instead of keeping different folders, we keep different docker images in the repositories. (it makes sense to change `:keep_releases` to 1). And don't have to redownload all gems, just like vanilla SSH deployment.

3. With the images in the repo, we can easily "cloud-deploy" the newly created image to clusters, for example with [ecs_deploy](https://github.com/reproio/ecs_deploy)

PR #370, opened 2016-09-19
Run default env variables sequentially instead of running in parallel.

xurumelous:
Sometimes the command starts before exporting the environment variables, crashing the command that depends on those variables.

### DIFF
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -163,7 +163,7 @@ module SSHKit
 
     def with(&_block)
       return yield unless environment_hash.any?
-      sprintf("( export #{environment_string} ; %s )", yield)
+      sprintf("( export #{environment_string} && %s )", yield)
     end
 
     def user(&_block)
--- a/test/functional/backends/test_netssh.rb
+++ b/test/functional/backends/test_netssh.rb
@@ -38,7 +38,7 @@ module SSHKit
           "Command: /usr/bin/env ls -l\n",
           "Command: if test ! -d /tmp; then echo \"Directory does not exist '/tmp'\" 1>&2; false; fi\n",
           "Command: if ! sudo -u root whoami > /dev/null; then echo \"You cannot switch to user 'root' using sudo, please check the sudoers file\" 1>&2; false; fi\n",
-          "Command: cd /tmp && ( export RAILS_ENV=\"production\" ; sudo -u root RAILS_ENV=\"production\" -- sh -c '/usr/bin/env touch restart.txt' )\n"
+          "Command: cd /tmp && ( export RAILS_ENV=\"production\" && sudo -u root RAILS_ENV=\"production\" -- sh -c '/usr/bin/env touch restart.txt' )\n"
         ], command_lines
       end
 
--- a/test/unit/test_command.rb
+++ b/test/unit/test_command.rb
@@ -28,44 +28,44 @@ module SSHKit
     def test_including_the_env
       SSHKit.config = nil
       c = Command.new(:rails, 'server', env: {rails_env: :production})
-      assert_equal %{( export RAILS_ENV="production" ; /usr/bin/env rails server )}, c.to_command
+      assert_equal %{( export RAILS_ENV="production" && /usr/bin/env rails server )}, c.to_command
     end
 
     def test_including_the_env_with_multiple_keys
       SSHKit.config = nil
       c = Command.new(:rails, 'server', env: {rails_env: :production, foo: 'bar'})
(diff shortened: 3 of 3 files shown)


klaussilveira:
:+1: this would solve some intermittent issues we've been experiencing.

bycosta:
Wow, good catch! Thanks

leehambley:
Thanks for the PR @xurumelous !

Could you please address the [Rubocop violation that broke the build](https://travis-ci.org/capistrano/sshkit/jobs/161087289#L447-L449) and address the contribution guidelines about crediting yourself and writing a proper changelog entry?

Related, it seems like the behaviour of `;`  vs `&&` might be doing something different than what most people expect in the shell.. what lead you to this conclusion and fix and are you sure there's a semantic difference, not just an incidental timing difference in `&&` blocking slightly longer for a return status or something?

mattbrictson:
Unrelated and at the same time, I happened to create a PR that fixes the RuboCop violation. #371. Feel free to cherry-pick my commit. 2cb0fd132a3734943946e60de921e35b90f6c9f1

Or we can merge mine first and you can rebase this PR. @leehambley?

leehambley:
No strong opinion, I'd just like to see some docs on the behaviour of ; vs.
&& my understanding is that both wait for the command to complete, one
however ignores the exit status, hence my suspicion that this isn't a
perfect fix.

robd:
I don't have any specific knowledge here, but I was interested to dig into this. Which shell are you using?

From https://www.gnu.org/software/bash/manual/bashref.html#index-commands_002c-lists

> Commands separated by a ‘;’ are executed sequentially; the shell waits for each command to terminate in turn. The return status is the exit status of the last command executed.

Could you share more info about how you came to the conclusion that the env vars were being created in parallel with the command being executed? Could it be that the creation of the environment vars is failing for some reason?

xurumelous:
That's true, both `;` and `&&` are sequentially.

We are using bash.

We are getting this issue for months now, and last week it became really annoying, since we could not deploy our application. The issue is intermittent, so it's being very hard to find the root cause. With this change, it worked right after many failing attempts. But, it just happened again, even with this change.

This is what happens:

```
DEBUG [0e6d70fe] Command: ( export COUNTRY="us" SYMFONY_ENV="prod" && /usr/bin/env php /foo/bar/releases/20160919182154/app/console cache:warmup  --no-debug --no-ansi )

DEBUG [0e6d70fe] 	PHP Fatal error:  Uncaught Exception: Environment variable "STORE" must be set. in /foo/bar/releases/20160919182154/app/AppKernel.php:35
```

The variable declaration:

```
set :default_env, {
  COUNTRY: :us
}
```

Ideas are very welcome :)

mattbrictson:
@xurumelous it looks like the app is expecting `STORE` to be set, but you haven't shown us where you are setting it. It could be as simple as adding it to `:default_env` like you've already done with `COUNTRY`...?

will-in-wi:
This PR is simply causing the subsequent commands to run when `export` is returning a non-0 status code.

I took a look at the bash docs on `export`, and they say that:

> export returns an exit status of 0 unless an invalid option is encountered, one of the names is not a valid shell variable name, or -f  is  supplied with a name that is not a function.

Is it possible that there are some weird characters in the name of the variable being exported?

Per the difference between `&&` and `;`:
* https://stackoverflow.com/questions/25669540/what-is-the-difference-between-double-ampersand-and-semicolon-in-linux
* https://superuser.com/questions/619016/what-is-the-difference-between-command-command-and-command-command/619019#619019

It looks like there is assumed left to right ordering, which would make these two equivalent, given a 0 exit status.

leehambley:
To support @will-in-wi's claim that this doesn't (shouldn't) affect behaviour:

http://tldp.org/LDP/abs/html/ops.html lists documentation for `&&` but makes no mention of `;` (makes sense, it's not an operator), however https://www.gnu.org/software/bash/manual/bashref.html#Lists explains how lists of commands work in Bash.

To quote the interesting line (note, no qualification about a change in behaviour in POSIX mode)

> A list is a sequence of one or more pipelines separated by one of the operators `;`, `&`, `&&`, or ‘||’, and optionally terminated by one of `;`, `&`, or a newline.

Also when Bash is run in POSIX mode (called via `/bin/sh`) https://www.gnu.org/software/bash/manual/html_node/Bash-POSIX-Mode.html claims no difference in behaviour of the `export` builtin. (see point 38, for the closest mentioned change to the behaviour of export)

xurumelous:
In the end looks it is a PHP issue while reading the environment variables. I'm closing the PR. Thanks for the support, guys.

PR #371, opened 2016-09-19
Fix "unnecessary splat" violation in RuboCop 0.43

mattbrictson:
Upgrading to RuboCop 0.43.0 revealed a new lint warning about an "unnecessary splat operator". I've fixed it.


mattbrictson:
Thanks for review! 🙇

PR #372, opened 2016-09-28
Use cp_r in local backend with recursive option

okuramasafumi:
When `recursive` optionfor `upload!` method is true, `local` argument is likely to be a String representing a directory, but `FileUtils.cp` doesn't accept it. So here we use `FileUtils.cp_r` instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ appear at the top.
 
   * Your contribution here!
   * [#390](https://github.com/capistrano/sshkit/pull/390): Properly wrap Ruby StandardError w/ add'l context - [@mattbrictson](https://github.com/mattbrictson)
+  * [#372](https://github.com/capistrano/sshkit/pull/372): Use cp_r in local backend with recursive option - [@okuramasafumi](https://github.com/okuramasafumi)
 
 ## [1.12.0][] (2017-02-10)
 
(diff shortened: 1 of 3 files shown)


leehambley:
Thanks for the patch @okuramasafumi we'll need a test to verify this as working, and as you're working in this area, you may want to fix the build, which is currently failing because `_options` shouldn't be prefixed if it's now used in the function according to Rubocop.

okuramasafumi:
@leehambley OK, I'll change variable name from `_options` to `option`. Also, I'll add tests for `upload!` method, which currently doesn't exist.

okuramasafumi:
@leehambley Btw, do you have a plan for a `test` or `stub` backend? I used local backend for tests, but it feels like it lacks some features, such as stubbing executions.

leehambley:
You ought to be able to use the [stubbing from the test framework][1], it's a fairly simple unit test that the method was *called*, I don't think we need to test the standard library behaves correctly!

[1]: https://github.com/seattlerb/minitest#stubs

okuramasafumi:
Sorry for the delay, I fixed the test so please check this again!

mattbrictson:
Hi @okuramasafumi, thanks again for the PR! I don't see the test that you mentioned. Can you make sure all your changes have been pushed?

okuramasafumi:
@mattbrictson I added tests and squashed commits.

mattbrictson:
Excellent work! Thanks ✨

PR #376, opened 2016-10-26
Extract host parser initialization into a separate method

vassilevsky:
Trying to improve the method as suggested here:

https://codeclimate.com/github/capistrano/sshkit/lib/sshkit/host.rb

### DIFF
--- a/lib/sshkit/host.rb
+++ b/lib/sshkit/host.rb
@@ -29,23 +29,7 @@ module SSHKit
         @hostname = "localhost"
         @user = ENV['USER'] || ENV['LOGNAME'] || ENV['USERNAME']
       elsif !host_string_or_options_hash.is_a?(Hash)
-        suitable_parsers = [
-          SimpleHostParser,
-          HostWithPortParser,
-          HostWithUsernameAndPortParser,
-          IPv6HostWithPortParser,
-          HostWithUsernameParser,
-        ].select do |p|
-          p.suitable?(host_string_or_options_hash)
-        end
-
-        if suitable_parsers.any?
-          suitable_parsers.first.tap do |parser|
-            @user, @hostname, @port = parser.new(host_string_or_options_hash).attributes
-          end
-        else
-          raise UnparsableHostStringError, "Cannot parse host string #{host_string_or_options_hash}"
-        end
+        @user, @hostname, @port = first_suitable_parser(host_string_or_options_hash).attributes
       else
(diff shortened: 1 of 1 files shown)


lib/sshkit/host.rb
```diff
@@ -94,6 +78,13 @@ def properties
       @properties ||= OpenStruct.new
     end
 
+    def first_suitable_parser(host)
+      unless (parser = PARSERS.find{|p| p.suitable?(host) })
+        fail UnparsableHostStringError, "Cannot parse host string #{host}"
+      end
```
mattbrictson: I think this logic would be easier to understand if the `parser` assignment was broken out onto its own line instead of being within the `unless` condition. What do you think of this?

```ruby
parser = PARSERS.find { |p| p.suitable?(host) }
fail UnparsableHostStringError, "Cannot parse host string #{host}" if parser.nil?
```

lib/sshkit/host.rb
```diff
@@ -192,4 +183,12 @@ def hostname
     end
   end
 
+  PARSERS = [
+    SimpleHostParser,
+    HostWithPortParser,
+    HostWithUsernameAndPortParser,
+    IPv6HostWithPortParser,
+    HostWithUsernameParser,
+  ]
```
mattbrictson: Maybe add `.freeze` here?

lib/sshkit/host.rb
```diff
@@ -29,23 +29,7 @@ def initialize(host_string_or_options_hash)
...
-        end
-
-        if suitable_parsers.any?
-          suitable_parsers.first.tap do |parser|
-            @user, @hostname, @port = parser.new(host_string_or_options_hash).attributes
-          end
-        else
-          raise UnparsableHostStringError, "Cannot parse host string #{host_string_or_options_hash}"
-        end
+        @user, @hostname, @port = first_suitable_parser(host_string_or_options_hash).attributes
```
mattbrictson: Nice improvement. 👍 I like the `first_suitable_parser` method name.

lib/sshkit/host.rb
vassilevsky: I asked Elsa and she kindly agreed to do it for us ☃️

mattbrictson:
Looks good, thanks! 🙇

vassilevsky:
CodeClimate GPA is now 3.94 ✨

PR #377, opened 2016-11-07
Extract ssh channel callbacks into separate methods

vassilevsky:
Deep nesting of blocks is hard to read and understand.

I have extracted them into several small methods:

* Small individual methods called `handle_EVENT` for each type of event
  that is processed by the library.

* The main setup method `set_handlers` that only sets channel callbacks.
  It does not try to handle the events itself but rather just calls
  the corresponding handler method for each callback.

  In other words, it acts as an event dispatcher.

I believe that this structure is well-maintainable:

* If it is necessary to add a new event handler, one could add a new
  `handle_EVENT` method and a new line to the `set_handlers` method,
  not a whole new block to `execute_command`.

* If it is necessary to change how an event is handled, one could only
  change the corresponding `handle_EVENT` method.

Unfortunately, it was necessary to introduce a whole object with
a method in place of a simple local variable `exit_status`
because only then it was possible to set the status inside
the `handle_exit_status` method.

The new `result` object might hold other values in the future though,
set in a similar way, so that might count as a future-proof thing.

I have deleted commented out code and empty blocks. I'm going to create issues for them instead if you greenlight this.

### DIFF
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -104,50 +104,40 @@ module SSHKit
       def execute_command(cmd)
         output.log_command_start(cmd)
         cmd.started = true
-        exit_status = nil
+        result = Struct.new(:exit_status).new(nil)
         with_ssh do |ssh|
           ssh.open_channel do |chan|
             chan.request_pty if Netssh.config.pty
-            chan.exec cmd.to_command do |_ch, _success|
-              chan.on_data do |ch, data|
-                cmd.on_stdout(ch, data)
-                output.log_command_data(cmd, :stdout, data)
-              end
-              chan.on_extended_data do |ch, _type, data|
-                cmd.on_stderr(ch, data)
-                output.log_command_data(cmd, :stderr, data)
-              end
-              chan.on_request("exit-status") do |_ch, data|
-                exit_status = data.read_long
-              end
-              #chan.on_request("exit-signal") do |ch, data|
-              #  # TODO: This gets called if the program is killed by a signal
-              #  # might also be a worthwhile thing to report
-              #  exit_signal = data.read_string.to_i
-              #  warn ">>> " + exit_signal.inspect
-              #  output.log_command_killed(cmd, exit_signal)
-              #end
-              chan.on_open_failed do |_ch|
-                # TODO: What do do here?
-                # I think we should raise something
-              end
-              chan.on_process do |_ch|
-                # TODO: I don't know if this is useful
-              end
-              chan.on_eof do |_ch|
-                # TODO: chan sends EOF before the exit status has been
-                # writtend
-              end
-            end
+            chan.exec(cmd.to_command) { set_handlers(chan, cmd, result) }
             chan.wait
           end
           ssh.loop
         end
         # Set exit_status and log the result upon completion
-        if exit_status
-          cmd.exit_status = exit_status
+        if result.exit_status
+          cmd.exit_status = result.exit_status
           output.log_command_exit(cmd)
         end
+      end
+
+      def set_handlers(channel, command, result)
+        channel.on_data                   { |*args| handle_data          command, *args }
+        channel.on_extended_data          { |*args| handle_extended_data command, *args }
+        channel.on_request('exit-status') { |*args| handle_exit_status   result,  *args }
+      end
+
+      def handle_data(command, channel, data)
+        command.on_stdout(channel, data)
+        output.log_command_data(command, :stdout, data)
+      end
+
+      def handle_extended_data(command, channel, _type, data)
+        command.on_stderr(channel, data)
+        output.log_command_data(command, :stderr, data)
+      end
+
+      def handle_exit_status(result, _channel, data)
+        result.exit_status = data.read_long
       end
 
       def with_ssh(&block)


leehambley:
Thanks for the PR @vassilevsky 

> Deep nesting of blocks is hard to read and understand.

You're referring to the code you removed, and refactored here, which is nearly only ever seen by people working on the SSHKit backend?

I'm curious what the motivation is here, it seems like a somewhat arbitrary change in code, without any functional differences (indicated by there being no test changes). Is this a precursor to a change you wanted to make, but couldn't because.

Personally I don't see a reason to merge this, the original code is as it always is Net::SSH examples, and is idiomatic considering our sister library.

I'll wait for feedback from the other maintainers, but this looks like a change for the sake of change, whilst the cynic in me thinks this might just be a way for you to register a daily FOSS contribution and keep some commit streak running to cultivate your GitHub commits graph.

I'm open to anything, if the other maintainers think this is a valuable change, or that we should accept it even if there's no perceptible gain because it was offered.

mattbrictson:
I think this is coming from a good place. We have a large chunk of nested code, and in general something like that can be improved by refactoring into small, well-named methods. So the intention is right. I appreciate your effort, @vassilevsky, and thanks for the PR.

However in this case, I think the refactor does not help much. Most of the benefit comes from deleting the commented out/unused code. The methods that are extracted all are coupled via a common shared state (channel, result, etc.) that other methods of the class do not have. That means perhaps extracting them into a single-purpose class might be a better approach (e.g. `CommandChannel` or some such).

Even so, as @leehambley pointed out: the existing pattern, although heavily-nested and involving large blocks (normally undesirable in Ruby code), is the prescribed usage of the net-ssh DSL. Therefore unless there is a really large benefit, we should probably not deviate from the standard practice.

As it stands I don't think I would merge this PR, but I would be on board with deleting the commented/unreachable code as an incremental improvement.

robd:
While I don't have a strong opinion in general about nested code, I'm afraid that I think this refactor makes the code harder to understand. I am happy to provide more feedback on the specifics if that would be helpful.

However I would like to understand more about what aspect of the nesting you found hard to read and understand @vassilevsky. Were you trying to add some new functionality or debug some existing functionality? Could you share what led you to dig into this code in the first place?

I'd like to add my experience of working in this code last year, when I was adding the support for interactive commands. I was coming at it from no experience of the netssh API, and I thought that it was pretty clear that the callbacks from netssh were being handled and passed into the SSHKit domain. I like the fact that the stdout and stderr cases are side by side and it's possible to see the similarities and differences between them at a glance.

I found the commented out code helpful because it showed me that at least the other possible netssh callbacks (eg `on_open_failed`, `on_process`, `on_eof`) had been considered to some extent, even though they are not handled. I would not be in favour of creating github issues for these cases unless there is a commitment to work on them in the near future. As far as I can see, that commented out code has been in since https://github.com/capistrano/sshkit/commit/def3df1cfa6612d60741c9ca940cfab70d3678d2 (3.5 years+), so I'm not sure if there is a case for handling them at all. I would not be against deleting that code if the consensus is that it's just noise.

I also have some thoughts about "refactor only" PRs. My preference is to refactor code as a precursor to some improvement in functionality (eg new feature / bug fix / improved test coverage). I see 2 risks with refactoring:

1) the new code shape might not support the next change which is needed in that area of the code
2) a bug might be introduced if the refactor accidentally changed some behaviour

Even though I really enjoy the refactoring process, I prefer to only take these risks and do refactoring in support of a change in functionality.

vassilevsky:
Gentlemen, thank you so much for taking the time to review my changes. I appreciate it. Your responses are overwhelmingly useful.

I've made the attempt to refactor this method because CodeClimate points it out:

https://codeclimate.com/github/capistrano/sshkit/lib/sshkit/backends/netssh.rb

(Switch to the **Source** tab for a more graphic illustration.)

This goes in line with Ruby Style Guide recommendation:

https://github.com/bbatsov/ruby-style-guide#three-is-the-number-thou-shalt-count

The reason I care about this is that I genuinely love this library. I use Capistrano daily, yet only recently got really interested in the low-level details of its operation. I have read the SSH RFC, net-ssh, and sshkit, and I like it a lot. I think that it deserves a solid 4.0 CC GPA.

I don't care about my GitHub streak one bit.

I realize that these changes are not perfect and solve `execute_command` complexity by creating some other complexity. Those nested blocks do have shared state that ain't easy to untangle.

I am perfectly fine with you closing this without merging. It was a great lesson already.

I also thought about extracting some classes but the thought was quite vague. It would be cool to hear Matt and Rob's ideas on this.

I might make another attempt at simplifying this area of code if I have time. Or maybe there are more important issues I could help with?

leehambley:
Thanks everyone for the constructive feedback, and @vassilevsky for having a thick-skin against my somewhat impolite comments. 

I'm not invested in either direction merging or not, though I concede that @robd has had more recent, deep contact with the internals than I have. 

Should we merge it, or not?

vassilevsky:
I have worked in tech support for 5 years 😉

robd:
Personally I wouldn't merge this PR, but with that in mind and since @vassilevsky asked for more feedback, I've added a few more thoughts which I hope may be helpful.

Thanks for sending across the code climate report. I took a look, and the errors are that there are TODOs, and that that the method is too long by 3 lines. I don't think that these 'bad score' markers are helpful, and I caution against trying to improve the code climate score in isolation. There is nothing inherently wrong with a 33 line method, or with TODOs, it depends entirely on what the TODOs are or what the 33 lines are. 

I've tried to show the approach I would take for refactoring this method if I was to limit my changes just to this class.

I have already discussed the TODOS and associated netssh callback blocks but I think these do add some noise and I think deleting them would help make it clearer. If someone needs to look into the other callbacks (`on_open_failed`, `on_process`, `on_eof`) I think it would be fine for them to have to go to the netssh documentation when the need arises. Deleting these would also solve the method length issue, and leaves the code in this state:

https://github.com/robd/sshkit/blob/75db3d0ced7d69e1ce8501c9767e00d877d05ecd/lib/sshkit/backends/netssh.rb#L109

Assuming we want to further improve readability here, the next thing I would think about is who will be reading this code and what aspect of the behaviour will they be interested in. I think it's probably going to be someone who is trying to understand how netssh interfaces to SSHKit. Therefore, I think that the most interesting lines of code are these:

https://github.com/robd/sshkit/blob/75db3d0ced7d69e1ce8501c9767e00d877d05ecd/lib/sshkit/backends/netssh.rb#L110-L122

Therefore I would try and keep this code at the top level so that people coming to the code don't have to dig through a method hierarchy to see this. If I wanted to reduce the complexity here, I would look to hide the netssh boiler plate code to do with running the loop:

https://github.com/robd/sshkit/blob/96105174df2fbc6197836b6802d43081eda3fa6a/lib/sshkit/backends/netssh.rb#L108

I think that this usefully hides the details of the syntax netssh needs to run a loop and removes this concern, which I imagine most people won't care about, from the top level. It also removes one of the layers of nesting.

I considered the following options for taking this further:

1) Push down setting pty and/or wait https://github.com/robd/sshkit/commit/65b26b7c6c8e6b340d516f25efceb4c740ea6d44

I rejected because I couldn't think of a good name for a method which does 3 things.

2) Extract handler methods for callbacks: https://github.com/robd/sshkit/commit/9431130c113dd01b9177cfd0e3926eb281479e62

I rejected this because I don't think there is enough useful encapsulation provided by these methods to justify having to dig into them. If the code in these blocks was more complex then I think this would be OK.

I hope this is of some help. There may be typos here because I'm not able to run the vagrant tests at the moment.

leehambley:
Thanks @robd - excellent feedback. I consider your comments the nail in the coffin for this PR, but I do think we should remove the comments and todos, and generally clean up, and consider the issue resolved.

If we're in agreement, @vassilevsky you might want to simply submit a PR (or change this one) to not do the method extraction, but rather just the simplification and we'll merge it.

vassilevsky:
Rob, thank you for your excellent feedback. The idea of extracting the loop rather than the callbacks is 💯!

Funnily enough, I started with the same change as you showed last. But I didn't stop there and took it further down the rabbit hole.

@mattbrictson could you also explain your idea about extracting a class?

In the meantime I am closing this without a single regret.

mattbrictson:
@vassilevsky I was thinking about a class like this. This might make testability somewhat easier (output handling and callback ordering can be exercised independent of a real SSH connection), but once you add all the necessary glue code to plug it into net-ssh, a lot of the readability will be lost and so this is probably a step backwards.

```ruby
class CommandExecutionObserver
  attr_reader :command, :output, :exit_status

  def initialize(command, output)
    @command = command
    @output = output
  end

  def received_stdout(channel, data)
    command.on_stdout(channel, data)
    output.log_command_data(command, :stdout, data)
  end

  def received_stderr(channel, _type, data)
    command.on_stderr(channel, data)
    output.log_command_data(command, :stderr, data)
  end

  def received_exit_status(_channel, data)
    self.exit_status = data.read_long
  end

  private

  attr_writer :exit_status
end
```

PR #378, opened 2016-11-24
do not prefix exec

dreyks:
`exec` is a shell builtin so `/usr/bin/env exec` results in `/usr/bin/env: exec: No such file or directory`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ appear at the top.
 
   * Add your entries below here, remember to credit yourself however you want
     to be credited!
+  * Do not prefix `exec` command
+    [PR #378](https://github.com/capistrano/sshkit/pull/378) @dreyks
 
 ## [1.11.4][] (2016-11-02)
 
--- a/lib/sshkit/command_map.rb
+++ b/lib/sshkit/command_map.rb
@@ -62,7 +62,7 @@ module SSHKit
 
     def defaults
       Hash.new do |hash, command|
-        if %w{if test time}.include? command.to_s
+        if %w{if test time exec}.include? command.to_s
           hash[command] = command.to_s
         else
           hash[command] = "/usr/bin/env #{command}"


leehambley:
Thanks for the patch!

Do you have a list of which shells have this as a prefix? I'm not sure it's
universal, maybe POSIX spec has more details?

If it's not 100% universal we can't merge this!

dreyks:
There's no binary named `exec`. just like there's no `test` binary. If some shell does not have such builtin it will fail regardless of whether it is prepended by `/usr/bin/env` or not

leehambley:
> There's no binary named exec. just like there's no test binary. 

> If some shell does not have such builtin it will fail regardless of whether it is prepended by /usr/bin/env or not

dreyks:
well, i don't completely understand why did you close this
right now if you use :exec it will get prepended by prefix and will fail for everyone.  with my change it will work for those shells that have `exec`

dreyks:
oh you updated your comment. i'll look at some shells and will give you examples

leehambley:
> oh you updated your comment. i'll look at some shells and will give you examples

Sorry, mis-clicked before I'd finished answering your feedback.

dreyks:
this is what i found

http://pubs.opengroup.org/onlinepubs/007904875/idx/sbi.html

not completely sure this is what you mean but seems like every posix-compliant shell has to implement these

leehambley:
Good find, according to Wikipedia:

> Originally, the name "POSIX" referred to IEEE Std 1003.1-1988, released in 1988. The family of POSIX standards is formally designated as IEEE 1003 and the international standard name is ISO/IEC 9945.

So the document you linked was correct.

leehambley:
Please update this PR with an entry in the CHANGELOG and to credit yourself, and I'll be happy to merge it. Thanks for your patience

dreyks:
:green_apple:

PR #383, opened 2016-12-26
Fix `Use assert_nil if expecting nil` Minitest warnings

mattbrictson:
This PR fixes the following Minitest warnings:

```
Use assert_nil if expecting nil from test/unit/test_host.rb:78:in `test_arbitrary_host_properties'. This will fail in MT6.
Use assert_nil if expecting nil from test/unit/test_command.rb:221:in `test_setting_exit_status'. This will fail in MT6.
```

Also fix a buggy test in `test_connection_pool.rb`.

### DIFF
--- a/test/unit/backends/test_connection_pool.rb
+++ b/test/unit/backends/test_connection_pool.rb
@@ -50,8 +50,8 @@ module SSHKit
       end
 
       def test_connections_are_reused_if_checked_in
-        conn1 = pool.with(connect, "conn") {}
-        conn2 = pool.with(connect, "conn") {}
+        conn1 = pool.with(connect, "conn") { |c| c }
+        conn2 = pool.with(connect, "conn") { |c| c }
 
         assert_equal conn1, conn2
       end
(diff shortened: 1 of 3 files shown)


leehambley:
Looks solid @mattbrictson, what's the story behind the:

``` diff
-        conn1 = pool.with(connect, "conn") {}
 -        conn2 = pool.with(connect, "conn") {}
 +        conn1 = pool.with(connect, "conn") { |c| c }
 +        conn2 = pool.with(connect, "conn") { |c| c }
```

?

leehambley:
Ohh, saw your commit message:

> When given an empty block, `pool.with` always returns `nil`, so the test
> was simply testing that `nil == nil`. Fix this by ensuring there are
> legitimate return values to compare.
> commit 756d5c2be798a875d6d9b1ea3b67b8ea9ab9c255

leehambley:
LGTM

mattbrictson:
Thanks for the quick review! 🎅

PR #386, opened 2017-02-06
Ability to specify default runner config

wcmonty:
The `SSHKit.config.default_runner` configuration option has the ability to set the default runner, but users cannot configure the default runner.  This adds the `SSHKit.config.default_runner_config` setting, that a user can use to configure the runner.

For example, a user might want to override the default group size of 2 in the following manner:

```ruby
SSHKit.config.default_runner = :groups
SSHKit.config.default_runner_config = { limit: 10 }
```

### DIFF
--- a/lib/sshkit/configuration.rb
+++ b/lib/sshkit/configuration.rb
@@ -3,7 +3,7 @@ module SSHKit
   class Configuration
 
     attr_accessor :umask
-    attr_writer :output, :backend, :default_env, :default_runner
+    attr_writer :output, :backend, :default_env, :default_runner, :default_runner_config
 
     def output
       @output ||= use_format(:pretty)
@@ -27,6 +27,10 @@ module SSHKit
 
     def default_runner
       @default_runner ||= :parallel
+    end
+
+    def default_runner_config
+      @default_runner_config ||= {}
     end
 
     def backend
--- a/lib/sshkit/configuration.rb
+++ b/lib/sshkit/configuration.rb
@@ -3,7 +3,7 @@ module SSHKit
   class Configuration
 
     attr_accessor :umask
-    attr_writer :output, :backend, :default_env, :default_runner, :default_runner_config
+    attr_writer :output, :backend, :default_env, :default_runner
 
     def output
       @output ||= use_format(:pretty)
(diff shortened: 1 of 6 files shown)


leehambley:
Hi @wcmonty looks good, thanks for the time and your contribution.

The only thing missing is some docs, somewhere (either EXAMPLES.md, or maybe the README?), most *new* users skip the CHANGELOG.

mattbrictson:
@wcmonty Can you also rebase on top of the latest `master` so that you pick up the CI changes I just made? This will ensure some automated checks are run against your PR. Thanks!

lib/sshkit/coordinator.rb
```diff
@@ -27,7 +27,7 @@ def each(options={}, &block)
     private
 
     def default_options
-      { in: SSHKit.config.default_runner }
+      { in: SSHKit.config.default_runner }.merge(SSHKit.config.default_runner_config)
```
robd: I was wondering, is it necessary to merge the `default_runner_config` here or could we just initialize `@default_runner_config` to `{ in: SSHKit.config.default_runner }` in sshkit/configuration.rb?

lib/sshkit/configuration.rb
```diff
@@ -29,6 +29,10 @@ def default_runner
       @default_runner ||= :parallel
     end
 
+    def default_runner_config
+      @default_runner_config ||= {}
```
robd: Perhaps it would be clearer to initialize this to `{ in: SSHKit.config.default_runner }` here?

wcmonty: @robd Thanks for your suggestions.  I added 8014f6d58ef3f136503df8fadfa9a706f3f5a23f, that uses the value of the `default_runner` as the value of the `:in` key.  This also allows a user to set the `default_runner` via the `default_runner_config`, although via the unintuitive `:in` key.

robd: @wcmonty Thanks very much for doing this. I hadn't really understood the complexity of what this would mean for the :in key in the config. I tried poking this around a bit, and I think this might be an improvement, although I might have missed something: https://github.com/robd/sshkit/commit/9c8714009c01fcf968511a366b27d4e50ac2482a

wcmonty:
@mattbrictson I went ahead and rebased off of the current `master`.

@leehambley I think that the README was a better spot for the docs.  Added in (c3561c660691afb081b19d8f07d78ee544bd1fd7).  Let me know if you would like me to add something to the `EXAMPLES.md`, although there isn't much in there currently in terms of runners.

It looks like `Danger` is failing the build, but I added the `* Your contribution here!` line back into  `Changelog.md`, so I'm not sure why.

mattbrictson:
Looks great, thanks! Sorry about the Danger mixup; I think its regex that checks the changelog has a mistake. This PR is good to go. 👍

PR #390, opened 2017-02-21
Properly wrap Ruby StandardError w/ add'l context

### LINKED ISSUE #389: Ruby StandardErrors (e.g. Errno::ECONNREFUSED)  should be wrapped by SSHKit with additional info, but aren't
SSHKit is supposed to wrap Ruby `StandardError`s in a new exception with additional context in the error message, such as host and user. This helps troubleshooting in e.g. Capistrano multi-host deploys. Without this context, it is hard to know where a command failed to execute.

SSHKit does this like so (taken from `SSHKit::Runner::Parallel`):

```ruby
begin
  backend(h, &block).run
rescue StandardError => e
  e2 = ExecuteError.new e
  raise e2, "Exception while executing #{host.user ? "as #{host.user}@" : "on host "}#{host}: #{e.message}"
end
```

Looks good, right? But no! The bug is that Ruby resolves the `StandardError` constant as `SSHKit::StandardError` which is something completely different than a Ruby `StandardError`. So Ruby errors like `Errno::ECONNREFUSED` go right past this `rescue` clause without getting matched and handled. Oops!

This bug was first reported here: https://github.com/capistrano/capistrano/issues/1848

mattbrictson:
Prior to this PR, any Ruby StandardError (such as socket connection errors) would not be rescued or interpreted by SSHKit. In a scenario where you are running commands on multiple hosts, this meant the error would be devoid of any context, like on which host the error occurred.

This behavior was actually a mistake: we were intending to rescue and wrap all Ruby StandardErrors, but the rescue clause was written such that it only caught `SSHKit::StandardError`, which is something entirely different.

Fix this by referring to the correct constant, and add unit tests.

Fixes #389.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#390](https://github.com/capistrano/sshkit/pull/390): Properly wrap Ruby StandardError w/ add'l context - [@mattbrictson](https://github.com/mattbrictson)
 
 ## [1.12.0][] (2017-02-10)
 
--- a/lib/sshkit/runners/parallel.rb
+++ b/lib/sshkit/runners/parallel.rb
@@ -10,7 +10,7 @@ module SSHKit
           Thread.new(host) do |h|
             begin
               backend(h, &block).run
-            rescue StandardError => e
+            rescue ::StandardError => e
               e2 = ExecuteError.new e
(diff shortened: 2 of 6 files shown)


will-in-wi:
This looks good to me!

mattbrictson:
Thanks! I'll merge it.

leehambley:
Late to the party, but this looks ace, nice one Matt!

Issue #369, opened 2016-09-07
SSHKit opens two connections instead of one when using SSHKit::Backend::Netssh::KnownHosts


mattbrictson:
This bug was first reported here, with the symptom of Capistrano prompting for the SSH password twice: https://github.com/capistrano/capistrano/issues/1774

This got me thinking that perhaps the SSHKit connection pool is actually creating two connections instead of one. And in fact, this is indeed the case.

I added a `puts` statement to `connection_pool.rb` like this:

``` diff
@@ -56,6 +56,7 @@ class SSHKit::Backend::ConnectionPool
   def with(connection_factory, *args)
     cache = find_cache(args)
     conn = cache.pop || begin
+      puts "*** Opening new SSH connection with args: #{args.to_s}"
       connection_factory.call(*args)
     end
     yield(conn)
```

And then ran `cap production deploy:check`. Sure enough, two connections were opened.

The interesting thing is that `args.to_s` generates two different results. Since `args.to_s` is used to generate a key for the connection pool, this is why the pool thinks it needs to create two separate connections.

`to_s` is changing because `SSHKit::Backend::Netssh::KnownHosts` has mutable state and is changing state in between the two SSH executions. You can see that in the second case, my `~/.ssh/known_hosts` has now been loaded and cached by `KnownHosts`, which changes its `to_s` output.

```
*** Opening new SSH connection with args: ["xx.xx.xx.xx", "deployer", {:user=>"deployer", :forward_agent=>true, :known_hosts=>#<SSHKit::Backend::Netssh::KnownHosts:0x007f9b78b0abb0 @_mutex=#<Thread::Mutex:0x007f9b78b0ab10>, @files={}>, :compression=>true, :keepalive=>true}]
.
.
.
*** Opening new SSH connection with args: ["xx.xx.xx.xx", "deployer", {:user=>"deployer", :forward_agent=>true, :known_hosts=>#<SSHKit::Backend::Netssh::KnownHosts:0x007f9b78b0abb0 @_mutex=#<Thread::Mutex:0x007f9b78b0ab10>, @files={"~/.ssh/known_hosts"=>#<SSHKit::Backend::Netssh::KnownHostsKeys:0x007f9b7a979b28 @_mutex=#<Thread::Mutex:0x007f9b7a979a88>, @path="/Users/mbrictson/.ssh/known_hosts", @hosts_keys={REDACTED...
```

I think we need to either:
- modify `KnownHosts#to_s` to be a stable value; or
[...]

PR #392, opened 2017-02-28
Fix open two connections with changed cache key

shirosaki:
`args.to_s` is used for cache key of connection pooling.
`ssh_options` in `args` is changed while creating connections in
`connection_factory.call(*args)`.
Resetting cache key using changed `args` prevents cache miss.

This is a workaround for #369.

### DIFF
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -54,6 +54,9 @@ class SSHKit::Backend::ConnectionPool
   # invoking the `connection_factory` proc with the given `args`. The arguments
   # are used to construct a key used for caching.
   def with(connection_factory, *args)
+    ssh_options = args.last
+    ssh_options_keys = ssh_options.keys if ssh_options.is_a? Hash
+    cache_key = args.to_s
     cache = find_cache(args)
     conn = cache.pop || begin
       connection_factory.call(*args)
@@ -61,6 +64,19 @@ class SSHKit::Backend::ConnectionPool
     yield(conn)
   ensure
     cache.push(conn) unless conn.nil?
+    update_cache_key(ssh_options, ssh_options_keys, cache_key, args)
+  end
+
+  # Update cache key with changed args to prevent cache miss
+  def update_cache_key(ssh_options, ssh_options_keys, cache_key, args)
+    return unless ssh_options_keys
+    args[args.size - 1] = ssh_options.reject { |k, _| !ssh_options_keys.include?(k) }
+    new_key = args.to_s
+    if cache_key != new_key
+      caches.synchronize do
+        caches[new_key] = caches.delete(cache_key)
+      end
+    end
   end
 
   # Immediately remove all cached connections, without closing them. This only
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -122,6 +122,7 @@ class SSHKit::Backend::ConnectionPool
 
     caches.synchronize do
       caches[new_key] = caches.delete(cache.key)
+      cache.key = new_key
     end
   end
 
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -54,9 +54,6 @@ class SSHKit::Backend::ConnectionPool
   # invoking the `connection_factory` proc with the given `args`. The arguments
   # are used to construct a key used for caching.
   def with(connection_factory, *args)
-    ssh_options = args.last
-    ssh_options_keys = ssh_options.keys if ssh_options.is_a? Hash
-    cache_key = args.to_s
     cache = find_cache(args)
     conn = cache.pop || begin
       connection_factory.call(*args)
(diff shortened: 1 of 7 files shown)


leehambley:
Hi @shirosaki thanks for the patch!

I'm really glad to see tests, and that you solved a bug. I'm not a 100% happy with the implementation *copying* changes into the cache keys.. maybe we could find another way to derive cache keys.

I'd be happy to merge this as-is (it has a test for correct behaviour) and eventually one-day re-evaluate (maybe) how we derive cache keys.

In short LGTM, I'm happy to merge if one of the other maintainers is!

mattbrictson:
Thanks for the PR!

I definitely think you are onto something here: the idea of re-computing the key after opening the connection sounds like a workaround I would be happy with.

However I have some concerns with the current implementation:

1. I believe the bug is due to option _values_ changing, not the keys. No new keys are added, and no keys are mutated. The issue is that one of the values is mutated. The fix (and the corresponding test) should be narrowly focused on this scenario.
2. If we simply add the connection to the cache with a recomputed key (i.e. without removing the original key) then we will have two references to the same connection in the cache. We should avoid this if possible. There are places in the code where we iterate over the connections with `cache.values`, and duplicate connections could lead to bugs.

Is there a way you could rework this PR to address these points?

Also, some nitpicks: 

* I think `args.last` would be preferable to the somewhat arbitrary `args[2]`.
* For code readability, I would like the "update cache key with changed args" logic to be factored into its own method with a self-documenting method name.

shirosaki:
Thanks for feedback.

The bug is due to option values changing and the cache key of `args.to_s` changing.
It would be nice if we could find new way to derive cache key.

I updated the patch for the following points.

- delete old key instead of just adding new key
- use `args.last`
- refactoring to a separated method

mattbrictson:
Thanks for the changes. I need some more time to review and think this over. It looks to me like this could be simplified even further, but I haven't figured it out yet.

mattbrictson:
I think there are a couple things I would change about this PR.

First, the test is not really an accurate reproduction of the original bug (#369). In the bug, the same SSH options hash is passed both times, and the hash contains the same keys both times. It is only the value of one the keys (`:known_hosts`) that gets mutated.

If we change the test to specifically cover only this scenario, then the code changes needed to make that test pass become a bit more simple: all we need to do is compare the key (i.e. `args.to_s`) that was used prior to `connection_factory.call` with the key that is generated after that call. If the key has changed (i.e. due to the options hash being mutated), then we update the cache.

Since this re-keying only needs to happen for the `Cache` and not for `NilCache` (i.e. when the connection pool is disabled), it makes sense to put the "key changed" logic on the cache object itself.

Anyway, I have pushed a commit to your branch with these proposed changes. Let me know what you think!

shirosaki:
Thanks for the commit. Indeed simplified.
But unfortunately the patch didn't solve password twice issue for my project using capistrano.

I did some fixes to avoid password twice.
https://github.com/shirosaki/sshkit/commit/43d6aa23dd9bfd3041e03496a8a9ab9d8f2ee0fc

`Host#netssh_options` creates new Hash from `ssh_options` when calling `with`. So new Hash would be required.
https://github.com/capistrano/sshkit/blob/master/lib/sshkit/host.rb#L66

:password_prompt and :logger options are added in net-ssh `Net::SSH.assign_defaults`.
The key in Cache object should be updated for comparing next time.

mattbrictson:
Yes, you are right. This bug goes much deeper than I originally realized. Thanks for pointing out that `:password_prompt` and `:logger` are being added by net-ssh; I didn't even know that was happening.

Maybe it is just me, it is difficult to follow what exactly is happening in the code, and as I've discovered it takes a lot of digging to understand why the workaround is written the way it is. I would therefore like to hold off on merging this PR.

Perhaps there is a better way to solve this? What if we opened a PR at net-ssh to change how they handle default options such that they make a copy of the options hash instead of modifying it in place? Would that make our workaround easier?

shirosaki:
I found another fix.
https://github.com/capistrano/sshkit/pull/392/commits/5a80b0d97b788542f8a2cb8e4420333b18a61efd
Set default options early in `SSHKit::Backend::Netssh`.
It is the same timing as `:known_hosts` default.

mattbrictson:
Very promising! This looks great. I'll review in more detail tomorrow. Thanks

lib/sshkit/backends/connection_pool.rb
```diff
@@ -103,11 +110,22 @@ def find_cache(args)
...
     end
   end
 
+  # Update cache key with changed args to prevent cache miss
+  def update_key_if_args_changed(cache, args)
+    new_key = cache_key_for_connection_args(args)
+    return if cache.same_key?(new_key)
+
+    caches.synchronize do
+      cache = caches[new_key] = caches.delete(cache.key)
```
mattbrictson: Is the chained assignment really needed here? I find it a bit hard to read.

If I remove the extra assignment and just write:

```ruby
caches[new_key] = caches.delete(cache.key)
cache.key = new_key
```

… then the tests still pass.

mattbrictson:
Awesome! ✨ 

I think this is good to merge, provided you can:

1. Rebase on master and fix the conflicts
2. Review my [code comment](https://github.com/capistrano/sshkit/pull/392#discussion_r106200280) above

Thanks for all your hard work on fixing this bug!

lib/sshkit/backends/connection_pool.rb
shirosaki: You are right. I'll fix it.

shirosaki:
Thanks for review.

- rebase on master and fix conflicts in CHANGELOG.md
- remove unneeded assignment

mattbrictson:
There is one functional test that is failing:

```
SSHKit::Backend::TestNetssh
  test_ssh_option_merge                                           FAIL (3.55s)
Minitest::Assertion:         --- expected
        +++ actual
        @@ -1 +1 @@
        -[:forward_agent, :known_hosts, :paranoid]
        +[:forward_agent, :known_hosts, :logger, :paranoid, :password_prompt]
        /Users/mbrictson/Code/sshkit/test/functional/backends/test_netssh.rb:62:in `test_ssh_option_merge'
```

If you are unable to run the functional tests locally, I can take care of it. Let me know!

shirosaki:
Fixed functional test.

BTW functional test causes many errors on Windows.

mattbrictson:
> BTW functional test causes many errors on Windows

Thanks for pointing that out. The functional tests don't work on Travis either, so there is definitely a lot of room for improvement!

mattbrictson:
Tests all pass. Thank you so much for this PR and all your research! 🙇

Issue #396, opened 2017-03-29
undefined method assign_defaults for Net::SSH:Module


cmiotto:
Last version 1.13.0 uses Net::SSH.assign_defaults that was introduced in net-ssh 4.0.0
You should update runtime dependency in gemspec

mattbrictson:
Yikes, good catch! We'll make this a top priority and push a fix this week.

@shirosaki do you have time to look into this? I would prefer to make our code work with net-ssh < 4.0.0 by doing a `respond_to?` check or something similar.

mattbrictson:
For reference, was introduced by https://github.com/capistrano/sshkit/pull/392.

mattbrictson:
@cmiotto Thank you very much for the report and for narrowing it down to the net-ssh dependency!

mattbrictson:
The fix is in master thanks to @shirosaki. I'll publish 1.13.1 tomorrow.

PR #397, opened 2017-03-30
Fix net-ssh default options with net-ssh < 4.0.0

shirosaki:
net-ssh < 4.0.0 doesn't have Net::SSH.aassing_defaults.
Fallback to logger assignment.
https://github.com/net-ssh/net-ssh/blob/v3.2.0/lib/net/ssh.rb#L211

Fix for #396 
I've confirmed `rake test:units` with net-ssh 3.2.0.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#397](https://github.com/capistrano/sshkt/pull/397): Fix NoMethodError assign_defaults with net-ssh older than 4.0.0 - [@shirosaki](https://github.com/shirosaki)
 
 ## [1.13.0][] (2017-03-24)
 
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -47,7 +47,16 @@ module SSHKit
 
         # Set default options early for ConnectionPool cache key
         def assign_defaults
-          Net::SSH.assign_defaults(@default_options)
+          if Net::SSH.respond_to?(:assign_defaults)
+            Net::SSH.assign_defaults(@default_options)
+          else
+            # net-ssh < 4.0.0 doesn't have assign_defaults
+            unless @default_options.key?(:logger)
+              require 'logger'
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#397](https://github.com/capistrano/sshkit/pull/397): Fix net-ssh default options with net-ssh < 4.0.0 - [@shirosaki](https://github.com/shirosaki)
```
mattbrictson: I would rephrase the changelog entry so that people know exactly what error is fixed. Something like this:

> Fix NoMethodError `assign_defaults` with net-ssh older than 4.0.0

lib/sshkit/backends/netssh.rb
```diff
@@ -47,7 +47,16 @@ def default_options
 
         # Set default options early for ConnectionPool cache key
         def assign_defaults
-          Net::SSH.assign_defaults(@default_options)
+          if Net::SSH.respond_to?(:assign_defaults)
+            Net::SSH.assign_defaults(@default_options)
+          else
+            # net-ssh < 4.0.0 doesn't have assing_defaults
```
mattbrictson: Typo: `assign_defaults`

shirosaki:
Thanks for review.
I've modified comments.

mattbrictson:
Great ✨
Thanks for the PR!!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#397](https://github.com/capistrano/sshkt/pull/397): Fix NoMethodError assign_defaults with net-ssh older than 4.0.0 - [@shirosaki](https://github.com/shirosaki)
```
reedloden: sshkit*

leehambley: I'm quite sure there's no typo here, are you certain there's an issue?

mattbrictson: Ah, I see there is a typo `sshkt` in the URL. I'll fix it in master. Good 👀 @reedloden

PR #399, opened 2017-04-26
Simplify code used to sanitize command

mattbrictson:
I was browsing the `command.rb` code and found a `sanitize_command!` method that was way more complicated than it needed to be. This PR refactors it to a one-liner. I also added a missing unit test.


mattbrictson:
I opted not to include a CHANGELOG entry since this is not a user-facing change.

mattbrictson:
@will-in-wi @leehambley is it worth merging this one?

will-in-wi:
I like this. Thanks!

Issue #133, opened 2014-04-26
Weird logging behaviour in upload! and download!


daniel-rikowski:
I've been experimenting with the `upload!` and `download!` methods and I noticed that their callback (`transfer_summarizer`) is outputting a lot of progress messages when processing large files.

For a 80 MB file I get roughly 5000 log lines at log level "debug" and still about 500 at log level "info".

1. I guess the number debug outputs are to expected, although personally I think logging every callback is a bit excessive.
2. Additionally the rule for logging info messages seems to be flawed:

		if percentage > 0 && percentage.round % 10 == 0
		  info message
		else
		  debug message
		end

It looks like the original intention is to output only 10% steps (0, 10, 20 ... 100%) but this code logs each and every callback when the percentage is near a 10% step.

For an 80 MB file I get:

	DEBUG Uploading ./target/package.tar.gz 89.42%
	DEBUG Uploading ./target/package.tar.gz 89.44%
	DEBUG Uploading ./target/package.tar.gz 89.46%
	DEBUG Uploading ./target/package.tar.gz 89.49%
	 INFO Uploading ./target/package.tar.gz 89.51%
	 INFO Uploading ./target/package.tar.gz 89.53%
	 INFO Uploading ./target/package.tar.gz 89.55%
	 INFO       ... about 40 more lines ...
	 INFO Uploading ./target/package.tar.gz 90.45%
	 INFO Uploading ./target/package.tar.gz 90.47%
	 INFO Uploading ./target/package.tar.gz 90.49%
	DEBUG Uploading ./target/package.tar.gz 90.51%
	DEBUG Uploading ./target/package.tar.gz 90.53%
	DEBUG Uploading ./target/package.tar.gz 90.55%

			
My guess is that `upload!` and `download!` are not intended and/or tested with large files.

I'm willing to improve the code, but I'm not sure where to start. 

As far as I understand the number of log messages starts to increase if the file size is about 100 times higher than the chunk size. (> 1,6 MB) because then we start to get callbacks with steps smaller than 1%.

As a test I changed the line 

        if percentage > 0 && percentage.round % 10 == 0

into			

        if percentage > 0 && percentage.round(digits) % 10 == 0

where `digits` increases with the number of total chunks. (`total / ch[:chunk_size]`) It greatly decreases the number of info logs, about 3-4 around the 10% steps but it that is a very dissatisfying solution.

[...]

PR #401, opened 2017-06-14
Add option to specify transfer log percentage

### LINKED ISSUE #334: Feature: make transfer_summarizer configurable
See discussion in #133

aubergene:
Fixes #334

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#401](https://github.com/capistrano/sshkit/pull/401): Add :log_percent option to specify upload!/download! transfer log percentage - [@aubergene](https://github.com/aubergene)
 
 ## [1.13.1][] (2017-03-31)
 
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -62,14 +62,14 @@ module SSHKit
       end
 
       def upload!(local, remote, options = {})
-        summarizer = transfer_summarizer('Uploading')
+        summarizer = transfer_summarizer('Uploading', options)
         with_ssh do |ssh|
           ssh.scp.upload!(local, remote, options, &summarizer)
         end
       end
 
       def download!(remote, local=nil, options = {})
-        summarizer = transfer_summarizer('Downloading')
+        summarizer = transfer_summarizer('Downloading', options)
         with_ssh do |ssh|
           ssh.scp.download!(remote, local, options, &summarizer)
         end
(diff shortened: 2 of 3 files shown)


aubergene:
The tests `test_upload`, `test_download` are mocked and don't actually test the real output anyway

mattbrictson:
Hi, thanks for the PR! I think we can go without a test in this case, but I still would like to understand how this feature is intended to be used. The original issue #133 is that `upload!` and `download!` are excessively verbose when dealing with large files. In place of an automated test, could you do a large file upload/download with this branch and show us the actual output that is produced? Thanks!

Also, there is a Travis failure due a RuboCop violation, but that is not related to this PR; you can disregard it.

However as @capistrano-bot [noted above](https://github.com/capistrano/sshkit/pull/401#issuecomment-308400833), I will need a CHANGELOG entry to merge this.

mattbrictson:
If you could also rebase on the latest `master`, that will resolve the RuboCop failure.

aubergene:
Ah, yes actually the fix for #133 already makes things a lot better (I only updated yesterday). The use case for us is that we are uploading to a minimum of 4 servers and as it's both fast and reliable having the Jenkins log filled with lots of uploading lines isn't very helpful, but actually ~40 lines isn't too bad.

Here's an example of it set to 25%

```ruby
upload! 'some_local_file.txt', '/home/some_user/somewhere'
  download! '/home/some_user/some_remote_file.txt', 'somewhere_local', :log_percent 25
```
screen shot 2017-06-15 at 11 34 14 am

I rebased to master, unsure why the Travis job failed for ruby 2.3.3

CHANGELOG.md
```diff
@@ -5,7 +5,7 @@ appear at the top.
 
 ## [Unreleased][]
 
-  * Your contribution here!
```
mattbrictson: Please don't remove this line! Thanks

lib/sshkit/backends/netssh.rb
```diff
@@ -96,14 +96,16 @@ def config
...
-      def transfer_summarizer(action)
+      def transfer_summarizer(action, options = {})
+        log_percent = options[:log_percent] || 10
+        log_percent = 100 if log_percent <= 0
         last_name = nil
         last_percentage = nil
         proc do |_ch, name, transferred, total|
           percentage = (transferred.to_f * 100 / total.to_f)
-          unless percentage.nan?
+          unless percentage.nan? || !log_percent
```
mattbrictson: Why would `log_percent` be falsey?

mattbrictson:
Thanks for the explanation and screen shot! I think this PR is good to merge after the code comments above are addressed.

aubergene:
Thanks, originally I was going to allow it to be `false` or a number, but it's fine just set it to 100 and you get a single event

mattbrictson:
Thanks! 🙇

PR #402, opened 2017-06-14
Fix RuboCop failure in master due to inappropriate shebang in Rakefile

mattbrictson:
Fixes this RuboCop violation:

```
Rakefile:1:1: W: Lint/ScriptPermission: Script file Rakefile doesn't have execute permission.
^^^^^^^^^^^^^^^^^^^
```

I also pinned the RuboCop version to prevent new releases of RuboCop from breaking the build.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,5 +1,3 @@
-#!/usr/bin/env rake
-
 require 'bundler/gem_tasks'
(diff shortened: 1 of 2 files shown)


will-in-wi:
Thanks, this makes a ton of sense.

PR #408, opened 2017-10-24
Teach upload! and download! to respect within(...)

sj26:
I've been using sshkit to automate some deployment tasks recently and found the tasks would be easier if `upload!` and `download!` respected my `within(...)` context.

They already seem to work with remote relative paths which use the remote cwd, but don't respect the `within(...)` path stack. But we can just prepend `pwd_path` to remote path if it's relative to make it work.

This might be a breaking change for some users, although the documentation does say to always use absolute paths.

Also I have no idea how to test this, beyond doing a whole bunch of manual testing against a real ssh remote. 😅

### DIFF
--- a/lib/sshkit/backends/local.rb
+++ b/lib/sshkit/backends/local.rb
@@ -11,6 +11,7 @@ module SSHKit
       end
 
       def upload!(local, remote, options = {})
+        remote = File.join(pwd_path, remote) unless remote[0] == "/"
         if local.is_a?(String)
           if options[:recursive]
             FileUtils.cp_r(local, remote)
@@ -25,6 +26,7 @@ module SSHKit
       end
 
       def download!(remote, local=nil, _options = {})
+        remote = File.join(pwd_path, remote) unless remote[0] == "/"
         if local.nil?
           FileUtils.cp(remote, File.basename(remote))
         else
--- a/lib/sshkit/backends/local.rb
+++ b/lib/sshkit/backends/local.rb
@@ -11,7 +11,7 @@ module SSHKit
       end
 
       def upload!(local, remote, options = {})
-        remote = File.join(pwd_path, remote) unless remote[0] == "/"
+        remote = File.join(pwd_path, remote) unless remote.start_with?("/")
         if local.is_a?(String)
           if options[:recursive]
             FileUtils.cp_r(local, remote)
@@ -26,7 +26,7 @@ module SSHKit
       end
 
       def download!(remote, local=nil, _options = {})
-        remote = File.join(pwd_path, remote) unless remote[0] == "/"
+        remote = File.join(pwd_path, remote) unless remote.start_with?("/")
         if local.nil?
           FileUtils.cp(remote, File.basename(remote))
         else
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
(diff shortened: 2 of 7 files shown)


sj26:
I totally missed the functional test directory. After some fun with Vagrant I got some tests similar to the existing tests which `upload!` into a nested `within()` structure to prove it all works. And I added a CHANGELOG entry.

mattbrictson:
Thanks for the PR! I appreciate the tests and CHANGELOG entry. I hope to have time to review this soon.

Would you say this is a breaking change?

sj26:
This doesn't break existing documented usage, which is to use absolute paths. That usage should be totally unaffected.

But if a user has relied on the incidental behaviour of relative paths being relative to the ssh-context's default current working directory (instead of sshkit's within(...) working directory) then that might break. But it was always documented that this wasn't supported, so maybe that's okay? Idk. Perhaps a minor version increment with a note calling out the change?

lib/sshkit/backends/local.rb
```diff
@@ -11,6 +11,7 @@ def initialize(_ = nil, &block)
       end
 
       def upload!(local, remote, options = {})
+        remote = File.join(pwd_path, remote) unless remote[0] == "/"
```
mattbrictson: Nitpick: I'd prefer `remote.start_with?("/")` instead of `remote[0] == "/"` here and elsewhere, for readability.

sj26: Oh yeah that's much nicer. 👍

I always avoid `#start_with?` thinking it was ActiveSupport or introduce in Ruby 2+, but it's been [around since 1.8.7](https://ruby-doc.org/core-1.8.7/String.html#method-i-start_with-3F). 🙈

sj26:
@mattbrictson Thanks for helping me get this through!

Totally agree with the nitpick, fixed. And I've added a note to the changelog which you can edit as you like. Also had to fix a bundle problem with Ruby 2.0 on Travis.

mattbrictson:
Looks good, thanks!

sj26:
Thanks! 🙇

When do you think this might make it into a release — is there anything else lined up?

mattbrictson:
This will probably go out on Friday or over the weekend

mattbrictson:
🚢 https://rubygems.org/gems/sshkit/versions/1.15.0

sj26:
🙇

Fjan:
Just had a script breaking because they relied on the old behaviour. Easy enough to fix, but as far as I can tell simply detecting that we're not in a within block and then keeping the new behaviour should be as easy as adding a check for `pwd_path.nil?`. So this:
`remote = File.join(pwd_path, remote) unless remote.start_with?("/") || pwd_path.nil?`
Perhaps we could add that as a refinement of this patch? Or alternatively return a more descriptive error than the current syntax error?

mattbrictson:
@Fjan Ah, you're right. Good catch! This breaks if you upload to a relative remote path without using a `within` block. Just curious: what is the expected behavior? It just uploads relative to the user's home directory?

Anyway, I agree that a `pwd_path.nil?` like you suggested is a good sanity check and will fix this bug. Would you be willing to submit a PR?

Fjan:
Yep, the old behaviour just uploads / downloads relative to the user's home directory. I'm a it swamped at work and have not contributed to the project before so it will take some time before I can get to doing a PR, so would be great if someone more experienced can do it.

Fjan:
Ok, I went ahead an created a PR (https://github.com/capistrano/sshkit/pull/411)

PR #411, opened 2017-11-18
Restore ability to work with a relative path outside a `within` block

Fjan:
This restores the behaviour of 1.14 for relative paths used in upload! / download! It fixes the issues introduced by #408 where scripts written for the old behaviour raise an exception


mattbrictson:
Looks good. Thanks! 🙌

PR #416, opened 2018-01-22
Lock version of rbnacl-libsodium for older versions of Ruby

will-in-wi:
Fixes this test failure: https://travis-ci.org/capistrano/sshkit/jobs/331658388


will-in-wi:
Note that an alternative solution is to deprecate older Ruby versions ourselves. However, this will have quite the affect on Capistrano and any other downstream applications.

mattbrictson:
Nice catch! 💯 

I agree, we need to be careful about dropping support for older Rubies ourselves. It will have to be done at some point, but perhaps with some more heads-up to affected users.

PR #417, opened 2018-01-24
Fix cache key generation performance for connection pooling

ElvinEfendi:
The PR https://github.com/net-ssh/net-ssh/pull/308/ introduced a new `known_hosts` options that sshkit uses as a default option: https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/netssh.rb#L38. It gets passed to `Net::SSH.start` at https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/connection_pool.rb#L59 through https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/netssh.rb#L176 as part of `args`. The same `args` is used to generate cache key for an established connection(https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/connection_pool.rb#L90). Once the connection is established the default known_hosts gets populated by `Net::SSH`. `know_hosts` file can include thousands of entries. That means every time `cache_key_for_connection_args` is called, Ruby has to do `to_s` on that huge object. This impacts performance for every single command executed.

As a fix the PR changes `Arrayto_s` to use `Array#hash` instead. The following shows the issue and impact of the change.

StackProf output without the changes in this PR:
```
> sshkit-test (v-1.15.1)$ bundle exec ruby main.rb
1.15.1
  INFO [6caeca9c] Running /usr/bin/env date on examplehost.com
args changed
  INFO [6caeca9c] Finished in 1.399 seconds with exit status 0 (successful).
  INFO [724692cb] Running /usr/bin/env date on examplehost.com
  INFO [724692cb] Finished in 0.211 seconds with exit status 0 (successful).
  INFO [d1a80064] Running echo done on examplehost.com
  INFO [d1a80064] Finished in 0.208 seconds with exit status 0 (successful).
> sshkit-test (v-1.15.1)$ stackprof tmp/stackprof.dump
==================================
  Mode: wall(1000)
  Samples: 1803 (0.66% miss rate)
  GC: 154 (8.54%)
==================================
     TOTAL    (pct)     SAMPLES    (pct)     FRAME
      1649  (91.5%)        1078  (59.8%)     SSHKit::Runner::Parallel#execute
       255  (14.1%)         255  (14.1%)     SSHKit::Backend::ConnectionPool#cache_key_for_connection_args
       154   (8.5%)         154   (8.5%)     (garbage collection)
       101   (5.6%)          90   (5.0%)     OpenSSL::PKey::EC.read_keyblob
[...]

PR #418, opened 2018-01-24
Use hostname username as cache key

ElvinEfendi:
This is an another solution to fix the issue described at https://github.com/capistrano/sshkit/pull/417. Please refer to the description of that PR for the details.

The assumption is `hostname` and `username` should be enough to identify a connection.

### DIFF
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -51,7 +51,7 @@ class SSHKit::Backend::ConnectionPool
 
   # Creates a new connection or reuses a cached connection (if possible) and
   # yields the connection to the given block. Connections are created by
-  # invoking the `connection_factory` proc with the given `args`. The arguments
+  # invoking the `connection_factory` proc with the given `args`. The first two arguments
   # are used to construct a key used for caching.
   def with(connection_factory, *args)
     cache = find_cache(args)
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -61,9 +61,6 @@ class SSHKit::Backend::ConnectionPool
     yield(conn)
   ensure
     cache.push(conn) unless conn.nil?
-    # Sometimes the args mutate as a result of opening a connection. In this
-    # case we need to update the cache key to match the new args.
-    update_key_if_args_changed(cache, args)
   end
 
   # Immediately remove all cached connections, without closing them. This only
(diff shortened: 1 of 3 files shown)


lib/sshkit/backends/connection_pool.rb
```diff
@@ -61,9 +61,6 @@ def with(connection_factory, *args)
     yield(conn)
   ensure
     cache.push(conn) unless conn.nil?
-    # Sometimes the args mutate as a result of opening a connection. In this
-    # case we need to update the cache key to match the new args.
-    update_key_if_args_changed(cache, args)
```
ElvinEfendi: This is not necessary anymore because hostname and username don't mutate as a result of opening a connection.

mattbrictson:
Thanks for the PR! I will try to elaborate more later, but in the meantime refer to my comment here for some background: https://github.com/capistrano/sshkit/issues/369#issuecomment-245966391

ElvinEfendi:
Thanks for the link @mattbrictson. It would be great if you can provide more details to the following excerpts from your comment there:

> The risk is that that 1% of users could see some really undesirable behavior, like a command being run on the wrong server in the worst case.

Assuming this PR(i.e using hostname and username only for cache key), how could that happen? Is there a real example where one would want to execute two commands on a given server with the same user but different ssh options?

> Further, we can't take a shortcut and assume the object ID (memory address) is a decent proxy for these objects, because Capistrano makes a deep copy of the entire host and netssh_options, so the memory addresses are constantly changing.

What about `#hash` as I have proposed in the other PR(https://github.com/capistrano/sshkit/pull/417)? Since it does change with the deep coping of objects and evidently it fixes the performance issue.

mattbrictson:
Username and password is not enough because people also other options to distinguish connections:

* port
* key
* proxy

I am sure there are others.

The `to_s` trick was a way to comprehensively consider all connection arguments, but apparently it is slow. I don't think we will find a solution that will cover 100% of cases.

But thank you for opening these PRs, because I do think this discussion is worth continuing.

One solution that this brings to mind: Perhaps we whitelist connection options that we know how to deal with:

* username, password, port, and key can trivially be concatenated into a string
* known_hosts can be ignored as it does not (?) contribute to distinguishing connections

If a connection consists only of these options, we can quickly and easily build an appropriate key.

However if there are options present that are not whitelisted (e.g. `:proxy`), then we disable connection pooling for that connection. Or we require the user specify `:connection_id` that can be used as the cache key for those cases.

Is that a good compromise?

mattbrictson:
Looks like I spoke too soon: #417 looks like a great alternative.

mattbrictson:
Closed in favor of #417

PR #419, opened 2018-02-08
[CI] Test against new Ruby 2.x versions

nicolasleger:


will-in-wi:
Seems good to me. @mattbrictson do you know if there is any reason for targeting the specific versions as we have been doing?

mattbrictson:
I think we targeted specific versions to guarantee we were getting the latest patch level. The aliases sometimes don't bring in the latest. For example, "2.4" is being resolved to 2.4.1 on Travis, even though the newest patch level is 2.4.3.

https://travis-ci.org/capistrano/sshkit/jobs/338934494

But, keeping our Travis config up to date to explicitly list the latest patches is also a lot of work, so we're probably going to be out of date one way or the other. This PR works for me.

Thanks @nicolasleger !

Issue #424, opened 2018-05-11
Bug in `as` group implementation


pblesi:
As far as I can tell, specifying group in an `as` block will not work due to a quoting bug.

Based on this [commit](https://github.com/capistrano/sshkit/commit/c464c9756ff56912c430291fc2bcb437401a56aa) of the group functionality, it seems that the escaping of the double quotes was necessary because the sudo command used double quotes. As of this [commit](https://github.com/capistrano/sshkit/commit/fbe177541288720e927292e4b7a6141031cd8a56#diff-4eaa2c55e6802df7b49a8e7a2f7584d5) single quotes for sudo are used and the double quotes for the command are incorrectly escaped.

If this is indeed correct, I can look to post a PR that fixes the issue, but I did not want to bother with a fix in case there is something I'm missing.

leehambley:
> Based on this commit of the group functionality, it seems that the escaping of the double quotes was necessary because the sudo command used double quotes.

Sounds totally plausible, what's the usecase for executing something `as` a group, is it a "hack" to make sure to set group ownershp/etc? I quoted "hack" because `as` was intended for users.

Looking forward to getting a plan

pblesi:
Yeah, I'm developing a DSL that leverages SSHKit and am trying to expose the various functionality it provides, so allowing users to pass group to the [as method](https://github.com/capistrano/sshkit/blob/master/lib/sshkit/backends/abstract.rb#L98-L116)
I don't have a specific use case outside of the one indicated in this [example](https://github.com/capistrano/sshkit/blob/master/EXAMPLES.md#run-a-command-with-a-different-effective-group-id)

I'll go ahead and open a PR to modify the functionality

mattbrictson:
Fixed via #425

PR #425, opened 2018-05-12
Fix escaped quotes bug in Command#group

pblesi:
Command#group incorrectly escapes double quotes, resulting in a syntax
error when specifying the group a command should be executed as. This
issue manifested when user command quotes [changed from double quotes to
single quotes](https://github.com/capistrano/sshkit/commit/fbe177541288720e927292e4b7a6141031cd8a56#diff-4eaa2c55e6802df7b49a8e7a2f7584d5). This fix removes the double quote escaping in order to prevent a syntax error when executing the command.

This addresses issue https://github.com/capistrano/sshkit/issues/424

### DIFF
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -182,7 +182,7 @@ module SSHKit
 
     def group(&_block)
       return yield unless options[:group]
-      "sg #{options[:group]} -c \\\"%s\\\"" % %Q{#{yield}}
+      "sg #{options[:group]} -c \"%s\"" % %Q{#{yield}}
       # We could also use the so-called heredoc format perhaps:
       #"newgrp #{options[:group]} <<EOC \\\"%s\\\" EOC" % %Q{#{yield}}
     end
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -182,7 +182,7 @@ module SSHKit
 
     def group(&_block)
       return yield unless options[:group]
-      "sg #{options[:group]} -c \"%s\"" % %Q{#{yield}}
+      %Q(sg #{options[:group]} -c "#{yield}")
       # We could also use the so-called heredoc format perhaps:
       #"newgrp #{options[:group]} <<EOC \\\"%s\\\" EOC" % %Q{#{yield}}
     end
--- a/test/unit/test_command.rb
+++ b/test/unit/test_command.rb
@@ -102,12 +102,12 @@ module SSHKit
 
     def test_working_as_a_given_group
       c = Command.new(:whoami, group: :devvers)
-      assert_equal "sg devvers -c \\\"/usr/bin/env whoami\\\"", c.to_command
+      assert_equal "sg devvers -c \"/usr/bin/env whoami\"", c.to_command
     end
 
     def test_working_as_a_given_user_and_group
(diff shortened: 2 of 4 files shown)


leehambley:
Could I get a second opinion on this please either @mattbrictson  or @will-in-wi it looks OK for me, but the escaped shell escaping thing always blows my mind.

will-in-wi:
I _think_ this makes sense. Shell escaping is tricky to follow.

lib/sshkit/command.rb
```diff
@@ -182,7 +182,7 @@ def umask(&_block)
 
     def group(&_block)
       return yield unless options[:group]
-      "sg #{options[:group]} -c \\\"%s\\\"" % %Q{#{yield}}
+      "sg #{options[:group]} -c \"%s\"" % %Q{#{yield}}
```
mattbrictson: Since we're already looking at this part of the code, I think there are some additional changes we can do to make this much easier to read.

First, I want to point out that:

* `%Q{#{yield}}` is the same as `"#{yield}"`
* `"#{yield}"` is the same as `yield.to_s`
* We always pass a string in the block, so `yield.to_s` is the same as `yield`

Taking this further:

* `"... %s ..." % yield` is the same as `"... #{yield} ..."`
* `"... \" ..."` can be written without escapes as e.g. `%Q(... " ...)`

And so this whole line could be much simplified to:

```ruby
%Q(sg #{options[:group]} -c "#{yield}")
```

test/unit/test_command.rb
```diff
@@ -102,12 +102,12 @@ def test_working_as_a_given_user
 
     def test_working_as_a_given_group
       c = Command.new(:whoami, group: :devvers)
-      assert_equal "sg devvers -c \\\"/usr/bin/env whoami\\\"", c.to_command
+      assert_equal "sg devvers -c \"/usr/bin/env whoami\"", c.to_command
```
mattbrictson: I think a single-quoted string would improve readability here.

```ruby
assert_equal 'sg devvers -c "/usr/bin/env whoami"', c.to_command
```

test/unit/test_command.rb
```diff
@@ -102,12 +102,12 @@ def test_working_as_a_given_user
...
     end
 
     def test_working_as_a_given_user_and_group
       c = Command.new(:whoami, user: :anotheruser, group: :devvers)
-      assert_equal "sudo -u anotheruser -- sh -c 'sg devvers -c \\\"/usr/bin/env whoami\\\"'", c.to_command
+      assert_equal "sudo -u anotheruser -- sh -c 'sg devvers -c \"/usr/bin/env whoami\"'", c.to_command
```
mattbrictson: `%Q` might help here, too.

```ruby
assert_equal %Q(sudo -u anotheruser -- sh -c 'sg devvers -c "/usr/bin/env whoami"'), c.to_command
```

leehambley:
Excellent feedback @mattbrictson thanks for doing the legwork to figure out how we can trip off four years of oxidation from this dark corner of the code.

pblesi:
Hi @mattbrictson Great feedback! I've incorporate the suggested changes, and things seem a lot more readable now. Mind taking another look?

mattbrictson:
Looks good, thanks! 🙌

I'll squash this down to one commit and merge the PR.

PR #426, opened 2018-05-21
Remove gemnasium badge

mattbrictson:
The Gemnasium badge on our README now says: `FIXME Migrate to GitLab`. This is because Gemnasium was acquired by GitLab and shut down: https://docs.gitlab.com/ee/user/project/import/gemnasium.html

I think solution is to just remove the badge.


leehambley:
Nice catch @mattbrictson

Issue #433, opened 2018-07-16
Redaction adds a space


NorseGaud:
Take the following for an example:

```
[2] pry(#<SSHKit::Backend::Netssh>)> execute :sudo, :bash, '-c "', :echo, "SERVICE_USER_PASSWORD=", redact("#{ENV['SERVICE_USER_PASSWORD']}"), ' >> /etc/systemd/system/mpx_micro_services_sidekiq.service.d/EnvironmentFile', '"'
11 sudo bash -c " echo SERVICE_USER_PASSWORD= [REDACTED]  >> /etc/systemd/system/mpx_micro_services_sidekiq.service.d/EnvironmentFile "
✔ 01 ctops_service_user@vams-ch2-1s.sys.XXXXX.net 0.146s => true
```

The result: `SERVICE_USER_PASSWORD= XXXXXXXXX`. Notice the space between = and X?

This looks to be the default behavior of sshkit, adding a space in between comma separated parameters to execute.

What do we think about forcing, when redaction is present, the removal of the space?

I tried including it as interpolation but see: 

```
[3] pry(#<SSHKit::Backend::Netssh>)> execute :sudo, :bash, '-c "', :echo, "SERVICE_USER_PASSWORD=#{redact(ENV['SERVICE_USER_PASSWORD'])}", ' >> /etc/systemd/system/mpx_micro_services_sidekiq.service.d/EnvironmentFile', '"'
FrozenError: can't modify frozen object
```

Curious what the capistrano team think about this problem?

mattbrictson:
Interesting find! I can't think of a good solution off the top of my head, but I am open to ideas.

> What do we think about forcing, when redaction is present, the removal of the space?

I think that behavior would be too surprising. E.g. removing a space before the password in this example would be strange:

```ruby
execute :dosomething, "--password", redact("password")
```

NorseGaud:
I’m thinking we can just get more explicit about the space and how someone should wrap the whole section if they want to prevent it. I’ll get the readme updated and in a pull request soon. I got a few examples from a recent project to add too.

NorseGaud:
https://github.com/capistrano/sshkit/pull/434

PR #434, opened 2018-07-23
[issues/433] New example for redaction readme

NorseGaud:
https://github.com/capistrano/sshkit/issues/433: Describing how to avoid the space messing up setting a env in linux.


NorseGaud:
Agreed. I made a new commit. What do you think?

NorseGaud:
Yes sir, just added them. Thanks for approving!

PR #435, opened 2018-07-26
Consistent verbosity configuration inside #capture and #test methods

NikolayRys:
Currently, the default params for those methods are applied differently, and I don't think that's intended. In the case of the #test method, the default params for `verbosity` and `raise_on_non_zero_exit` are overriding the provided by the user.
It makes impossible to have important #test calls show any output. At least without an ugly and non-thread safe `SSHKit.config.output_verbosity=Logger::DEBUG`.

So the proposed PR makes them consistent, using the approach from #capture and provides a test to cover that.
The overall change is pretty small.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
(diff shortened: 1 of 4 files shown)


leehambley:
LGTM anyone else want to have a look-in?

NikolayRys:
No luck with this, guys?

will-in-wi:
Sorry for the delay, this got lost in the shuffle.

will-in-wi:
And thanks for contributing!

Issue #438, opened 2018-08-10
ConnectionPool: idle_timeout of 0 causes 100% and busy looping


tisba:
I recently noticed that our sidekiq is running at 100% CPU (see https://github.com/mperham/sidekiq/issues/3918). We are using sshkit in our sidekiq workers.

We were using:

```ruby
SSHKit::Backend::Netssh.pool.idle_timeout = 0
```

Which causes a busy loop in `ConnectionPool#run_eviction_loop` as `[idle_timeout, 5].min` will be `0`:

https://github.com/capistrano/sshkit/blob/70d01df12b0f2db92baf086d14387ef677a32a38/lib/sshkit/backends/connection_pool.rb#L131-L139

Also, according to `ConnectionPool`s documentation, `idle_timeout` can also be `false` or `nil` which will actually cause an exception (`[nil, 5].min` will raise).

I'm happy to propose a fix, but I'd like to have guidance how. My proposal would be to impose a lower limit on `run_eviction_loop` loop frequency like at most once every 5 seconds. On the other hand I don't understand why the eviction loop is required at all if connection caching is disabled (`idle_timeout = 0`, `ConnectionPool#cache_enabled?` -> `false`).

As far as I can tell the code was heavily changed/rewritten with https://github.com/capistrano/sshkit/pull/328 by @mattbrictson. Maybe he can provide some guidance?

In any case, we can work with increasing the `idle_timeout` to `>0` for now. I still like to have this problem fixed though.

mikegee:
No help for the underlying issue, but I don't think you would have seen this problem if capistrano weren't loaded in the Sidekiq process. The docs say to use `require: false` in the `Gemfile`. https://github.com/capistrano/capistrano#install-the-capistrano-gem

tisba:
To clarify things: We do not load capistrano in sidekiq, only sshkit. We use sshkit directly to manage servers.

tisba:
-bump-

I'm happy to provide a patch to the problem (outlined in the description above). Is this a viable way to solve this issue?

leehambley:
Hi @tisba if you can provide a patch it would be most welcome. We are still actively maintaining this library when community patches are provided, but both @mattbrictson and myself are not spending a lot of time on bug fixes ourselves.

Please do submit a fix, if you can, and the doc issue:

> Also, according to ConnectionPools documentation, idle_timeout can also be false or nil which will actually cause an exception ([nil, 5].min will raise).

[...]

PR #448, opened 2019-01-29
Fix Disabling of Connection Pool

tisba:
Fixes https://github.com/capistrano/sshkit/issues/438

The README currently states, that:

```ruby
SSHKit::Backend::Netssh.pool.idle_timeout = 0
```

will disable the cache and this can be used to mitigate potential problems from connection pooling. As described in #438 quite the opposite was the case. Connections were not cached, BUT the cache eviction loop was still started. With `idle_timeout = 0` the the eviction loop becomes a busy loop hogging one core at 100%.

This PR will skip the launch of the eviction loop (via `run_eviction_loop`) in case the cache is disabled. If the cache is disabled, there can't be any `timed_out_connections` because `silently_close_connection_later` is not used.

`find_cache` does not use `thread_safe_find_or_create_cache` if cache is disabled:

https://github.com/capistrano/sshkit/blob/ac181607508de1f1dbc8cb3b6df963ed343f5418/lib/sshkit/backends/connection_pool.rb#L98-L106

I'd like to get some guidance on how to test this (@leehambley maybe?).

## Further Suggestion

I also would like to propose that there is a maximum eviction frequency. The `idle_timeout` should not guarantee when idle connections are removed. Rather it should be considered "after `idle_timeout` sshkit will evict the connection eventually; never before". If you agree with this, we could simplify the `sleep` from this:

https://github.com/capistrano/sshkit/blob/ac181607508de1f1dbc8cb3b6df963ed343f5418/lib/sshkit/backends/connection_pool.rb#L129-L139

to

```ruby
sleep 5
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@ appear at the top.
 
 ## [Unreleased][]
 
-  * Your contribution here!
+  * [#448](https://github.com/capistrano/sshkit/pull/448): Fix misbehaving connection eviction loop when disabling connection pooling - [Sebastian Cohnen](https://github.com/tisba)
 
 ## [1.18.1][] (2019-01-26)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@ appear at the top.
 
 ## [Unreleased][]
 
+  * Your contribution here!
   * [#448](https://github.com/capistrano/sshkit/pull/448): Fix misbehaving connection eviction loop when disabling connection pooling - [Sebastian Cohnen](https://github.com/tisba)
 
 ## [1.18.1][] (2019-01-26)
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -46,7 +46,11 @@ class SSHKit::Backend::ConnectionPool
     @caches = {}
     @caches.extend(MonitorMixin)
     @timed_out_connections = Queue.new
-    Thread.new { run_eviction_loop }
+
+    # Spin up eviction loop only if caching is enabled
+    if cache_enabled?
+      Thread.new { run_eviction_loop }
+    end
   end
 
   # Creates a new connection or reuses a cached connection (if possible) and
--- a/lib/sshkit/backends/connection_pool.rb
+++ b/lib/sshkit/backends/connection_pool.rb
@@ -21,7 +21,7 @@ end
 
 # The ConnectionPool caches connections and allows them to be reused, so long as
 # the reuse happens within the `idle_timeout` period. Timed out connections are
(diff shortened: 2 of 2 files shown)


leehambley:
@tisba thanks for the submission, this seems like a great fix. It's dirt simple, and the explanation in the PR is longer than the code sample, which inspires confidence.

I've requested a review from @mattbrictson if he's OK with it, let's roll it and get it out in the next release!

tisba:
Nice!

Do you want to me to update the changelog, as the nice @capistrano-bot mentioned? :)

Also, what about the other suggestion regarding the slight adjustment of the eviction loop as well?

leehambley:
> Do you want to me to update the changelog, as the nice @capistrano-bot mentioned? :)

Probably a good idea!

> Also, what about the other suggestion regarding the slight adjustment of the eviction loop as well?

That concerns me somewhat more than this, lets keep them separate? @mattbrictson what do you think?

mattbrictson:
> Also, what about the other suggestion regarding the slight adjustment of the eviction loop as well?

I think it is safe to include. Just to clarify, we're talking about changing this line:

```ruby
 sleep([idle_timeout, 5].min)
```

to be:

```ruby
sleep(5)
```

correct?

> There are code changes, but no corresponding tests. Please include tests if this PR introduces any modifications in behavior.

I am OK with accepting the PR without tests.

Thanks so much for this PR! 🙏

tisba:
This is exactly the change I would like to suggest.

With https://github.com/capistrano/sshkit/pull/448/commits/162cbfd503e039c1b2b9e2e39ad7659b22e88ab9 the eviction of timed out connection could obviously take a couple of seconds, but (like I already outlined) the semantics of "will eventually be closed/removed" is not something that users of sshkit should be surprised by.

leehambley:
Danke für den Einsatz! 

Thanks for your contribution, may your SideKiq workers spin less frequently on a busy loop! 🏆

mattbrictson:
We generally publish new SSHKit releases to rubygems every six weeks, and since the most recent release was just a few days ago, that would mean this fix doesn't get published until about 5 weeks from now. But I'm happy to make exceptions for important bug fixes. Does this qualify?

tisba:
IMO this is a nasty bug, even though sshkit users might run into it only extreme rare cases. What makes it "worse" IMO, is that the current README (regarding the ability to disable the connection cache) directly leads to the issue fixed by this PR.

On the other hand the mitigation (once you've figured out what is going on), is rather trivial: Don't disable the connection pool.

Me personally I'm okay with not rushing this.

mattbrictson:
Actually, I'll make a release today. These sorts of major bug fixes don't happen very often.

mattbrictson:
🚀 released as 1.18.2

PR #452, opened 2019-02-19
unify string replacement styles

grosser:


mattbrictson:
No need for a CHANGELOG entry on this one since it is not user facing.

Congrats on your first contribution to the project!

PR #453, opened 2019-02-19
fix and unify shell escaping for user/group/directory

grosser:
- support user/group/directories with spaces
- fixing the broken user/group escaping

... not touching env vars since I think someone might be abusing this for $ var injection, needs more testing

### DIFF
--- a/test/unit/test_command.rb
+++ b/test/unit/test_command.rb
@@ -51,13 +51,13 @@ module SSHKit
     def test_double_quotes_are_escaped_in_env
       SSHKit.config = nil
       c = Command.new(:rails, 'server', env: {foo: 'asdf"hjkl'})
-      assert_equal %{( export FOO="asdf\\\"hjkl" ; /usr/bin/env rails server )}, c.to_command
+      assert_equal %{( export FOO="asdf\\"hjkl" ; /usr/bin/env rails server )}, c.to_command
     end
 
     def test_percentage_symbol_handled_in_env
       SSHKit.config = nil
       c = Command.new(:rails, 'server', env: {foo: 'asdf%hjkl'}, user: "anotheruser")
-      assert_equal %{( export FOO="asdf%hjkl" ; sudo -u anotheruser FOO=\"asdf%hjkl\" -- sh -c '/usr/bin/env rails server' )}, c.to_command
+      assert_equal %{( export FOO="asdf%hjkl" ; sudo -u anotheruser FOO=\"asdf%hjkl\" -- sh -c /usr/bin/env\\ rails\\ server )}, c.to_command
     end
 
     def test_including_the_env_doesnt_addressively_escape
@@ -84,6 +84,11 @@ module SSHKit
       assert_equal "cd /opt/sites && /usr/bin/env ls -l", c.to_command
     end
 
+    def test_working_in_a_given_weird_directory
+      c = Command.new(:ls, '-l', in: "/opt/sites and stuff")
+      assert_equal "cd /opt/sites\\ and\\ stuff && /usr/bin/env ls -l", c.to_command
+    end
+
     def test_working_in_a_given_directory_with_env
       c = Command.new(:ls, '-l', in: "/opt/sites", env: {a: :b})
       assert_equal %{cd /opt/sites && ( export A="b" ; /usr/bin/env ls -l )}, c.to_command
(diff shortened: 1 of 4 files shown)


test/unit/test_command.rb
```diff
@@ -33,60 +33,65 @@ def test_leading_and_trailing_space_is_stripped
     def test_including_the_env
       SSHKit.config = nil
       c = Command.new(:rails, 'server', env: {rails_env: :production})
-      assert_equal %{( export RAILS_ENV="production" ; /usr/bin/env rails server )}, c.to_command
+      assert_equal %{( export RAILS_ENV=\"production\" ; /usr/bin/env rails server )}, c.to_command
```
mattbrictson: This seems incorrect to me. The backslashes mean that the quotes will be included in the value. For example if I run this locally:

```
$ export RAILS_ENV=\"production\" ; echo $RAILS_ENV
"production"
```

grosser: this is just a display problem, it is the correct string:

```
puts %{( export RAILS_ENV=\"production\" ; /usr/bin/env rails server )}
( export RAILS_ENV="production" ; /usr/bin/env rails server )
```

I removed the extra escapes from the assertion and it looks saner now

grosser:
added another commit to avoid calling env methods multiple times

grosser:
good to go ?

grosser:
@mattbrictson

grosser:
@mattbrictson is there anything missing ?

mattbrictson:
Sorry for the radio silence! To be honest I have not been able to find much time in recent weeks for Capistrano PR reviews and so I am prioritizing those that are low-risk, high-value, and well-documented.

This PR is tricky because it definitely changes existing behavior (as evidenced by the fact that you needed to change existing tests) and I don't want to merge without be absolutely certain all the bases are covered.

I would be more comfortable merging this if the PR description (or alternatively, a bug report linked to this PR) explained in more detail what problem this PR is solving with some examples of errors or unwanted behavior. Also, considering there is some risk of changing existing behavior, it would be nice to explain who might be affected by these changes, and what they could to do mitigate any problems when upgrading. A more detailed CHANGELOG entry that summarizes this would also help.

I know this might be discouraging but given the large number of users of Capistrano and the fact that the project support 100% by volunteers, I tend to be very conservative about merging PRs unless there are good supporting docs and migration paths.

grosser:
examples:

within
```Ruby
require 'sshkit'

include SSHKit::DSL

on 'server' do
  within 'space here' do
    puts capture "pwd"
  end
end

pwd stderr: bash: line 0: cd: space: No such file or directory
```

user part 1
```Ruby
on 'bastion1.use1.zdsystest.com' do
  as 'space here' do
    puts capture "echo 1"
  end
end

if ! sudo -u space here whoami > /dev/null; then echo "You cannot switch to user 'space here' using sudo, please check the sudoers file" 1>&2; false; fi stderr: sudo: unknown user: space
```

user part 2
```
on 'bastion1.use1.zdsystest.com' do
  as 'nobody' do
    puts capture :echo, "'"
  end
end

echo stderr: bash: -c: line 0: unexpected EOF while looking for matching `''
```

group part 1
```Ruby
on 'bastion1.use1.zdsystest.com' do
  as group: 'space here' do
    puts capture :echo, '1'
  end
end

echo stderr: sudo: no tty present and no askpass program specified
```

group part 2
```Ruby
on 'bastion1.use1.zdsystest.com' do
  as group: 'nobody' do
    puts capture :echo, '"'
  end
end

echo stderr: bash: -c: line 0: unexpected EOF while looking for matching `"'
```

Also in general, this is a gem that connects to maybe hundreds of live servers to execute commands, which is terrifying, it should be built on a solid foundation and not "this kinda seems to work, why change it", shellwords is a builtin library for exactly this usecase. Whatever shady code worked before because it "hacked this hack with another hack" is wrong & dangerous.

The changelog warning could read "if you replied on strange quoting to break out of the surrounding sh -c then this will no longer work" / "if you put env vars into usernames then we will escape them now" ... but afaik it should not break anything for anyone doing regular command execution.

grosser:
in general it seems like a bad idea to use these user/within apis anyways since they add the overhead of another remote execution (check user exists / check dir exists) to each execution.

mattbrictson:
I will clean up the CHANGELOG, fix the conflicts and get this merged in. Thanks for the examples. If you don't mind I will move them to their own issue to make them more discoverable.

I hope our future interactions on GitHub will not be so abrasive! 🙂🙏

grosser:
thx!

mattbrictson:
Merged via #459

grosser:
awesome! :D

I'd say let's get this out in a release and if it does not cause complaints within a month then revisit https://github.com/capistrano/sshkit/pull/451

PR #455, opened 2019-02-21
Ensure UUID of command is shared when redacted

### LINKED ISSUE #454: Commands and command output seem to get different UUID values
In the logs for our deployments, I often see output like this:

```
  INFO [01e05be1] Running /usr/bin/env git remote update --prune as user@hostname
 DEBUG [01e05be1] Command: cd /data/repo && ( export GIT_ASKPASS="/bin/echo" GIT_SSH="/tmp/git-ssh-app-env-username.sh" ; /usr/bin/env git remote update --prune )
 DEBUG [b183598d] 	Fetching origin
 DEBUG [b183598d] 	From github.com:organisation/app-repo
 x [deleted]         (none)     ->some_branch
 DEBUG [b183598d] 	 x [deleted]         (none)     -> some_other_branch
 DEBUG [b183598d] 	remote: Enumerating objects: 42, done.
remote: Counting objects:   2% (1/42)
remote: Counting objects:   4% (2/42)
remote: Counting objects:   7% (3/42) 
remote: Counting objects:   9% (4/42)
...
 DEBUG [b183598d] 	Unpacking objects:  95% (40/42)   
 DEBUG [b183598d] 	Unpacking objects:  97% (41/42)   
 DEBUG [b183598d] 	Unpacking objects: 100% (42/42) 
  INFO [b183598d] Finished in 4.171 seconds with exit status 0 (successful).
```

Ignoring the specific command, what is causing me some problems are the different command UUIDs that are appearing, so in this instance:

* `01e05be1` for the top-level command ("Running" / "Command" lines)
* `b183598d` for the output from the command (output and "Finished" lines)

The user-level problem here is that having different UUIDs makes it much harder for me to trace issues in deployments with multiple servers.

This seems to happen regardless of whether or not the deploy is to a single server or multiple servers (i.e. I see it in both types of Capistrano stage), and isn't specific to git commands (i.e. I see it in bundle installs and pretty much every other command that isn't trivial).

I've tried to replicate this in the simplest possible case so I can figure out why it's happening, but that's the weird part: I can't. Given this task:

```ruby
task :dummy_task do
  on roles([:solo, :app]) do
    within release_path do
      execute("/usr/bin/env ruby -v")
    end
  end
end
```

Here's the logging I see:
```
  INFO [1433932f] Running /usr/bin/env ruby -v as user@host
 DEBUG [1433932f] Command: /usr/bin/env ruby -v
 DEBUG [1433932f] 	ruby 2.4.5p335 (2018-10-18 revision 65137) [x86_64-linux-gnu]
  INFO [1433932f] Finished in 0.183 seconds with exit status 0 (successful).
```

... the same UUID for both the "Running" / "Command" lines, and for the output from the command. This seems to be the case regardless of the command that I put into the `execute` call. So unfortunately I don't seem to be able to generate a minimal replicable case ☹️ 

Do you have any idea why this might be happening? Is this expected behaviour? If I can at least get a minimal test case together, I'd be very happy to explore a fix.

---

Environment details:

sshkit (1.17.0)
capistrano (3.11.0)
capistrano-bundler (1.3.0)
capistrano-faster-assets (1.1.0)
capistrano-maintenance (1.2.0)
capistrano-rails (1.4.0)
airbrussh (1.3.0)

lazyatom:
In logging, the command UUID displayed in the "command start" phase was sometimes different to that displayed with the output of the command. This is because the backend can call `.with_redaction` on a command when logging the start, and this causes the command to be duplicated. However, if the `uuid` method hadn't been called before that point, the copy command would generate its own UUID, distinct from the original command.

This change avoids that by eagerly calculating the UUID when the command is created, so that any copied objects share the same UUID.

Fixes #454

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+* [#455](https://github.com/capistrano/sshkit/pull/455): Ensure UUID of commands are stable in logging - [@lazyatom](https://github.com/lazyatom)
 
 ## [1.18.2][] (2019-02-03)
 
--- a/lib/sshkit/command.rb
+++ b/lib/sshkit/command.rb
@@ -9,7 +9,7 @@ module SSHKit
 
     Failed = Class.new(SSHKit::StandardError)
 
-    attr_reader :command, :args, :options, :started_at, :started, :exit_status, :full_stdout, :full_stderr
+    attr_reader :command, :args, :options, :started_at, :started, :exit_status, :full_stdout, :full_stderr, :uuid
 
     # Initialize a new Command object
     #
@@ -25,6 +25,7 @@ module SSHKit
       @args    = args
       @options.symbolize_keys!
       @stdout, @stderr, @full_stdout, @full_stderr = String.new, String.new, String.new, String.new
+      @uuid = Digest::SHA1.hexdigest(SecureRandom.random_bytes(10))[0..7]
     end
 
     def complete?
@@ -39,10 +40,6 @@ module SSHKit
     def started=(new_started)
       @started_at = Time.now
       @started = new_started
-    end
-
-    def uuid
-      @uuid ||= Digest::SHA1.hexdigest(SecureRandom.random_bytes(10))[0..7]
     end
 
     def success?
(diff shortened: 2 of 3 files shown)


mattbrictson:
Excellent PR, thanks! Congratulations on your first code contribution to the project 🎉

leehambley:
Kudos @lazyatom really nice fix, practical and has flown under the radar for circa 5 years! 🥇

lazyatom:
My pleasure, and thanks for your help and work on this project! 🎉

PR #457, opened 2019-03-04
always close pool connections to avoid zlib warnings

eatwithforks:
previously, we only closed connections when there are no errors.
when we hit errors while executing, we saw misleading zlib errors where instead it was just the ssh command failing.
so lets always close connections to avoid those errors.

Fix for https://github.com/capistrano/sshkit/pull/343. 
@mattbrictson


grosser:
@mattbrictson this one would be nice too, unless there was a reason for the old code

grosser:
@mattbrictson 🛎

leehambley:
@grosser it's incredibly impolite to ping a contributor by name. Matt is taking some much needed personal time away from the project and @will-in-wi  and I are holding down the fort.

grosser:
- idk who else to ping
 - there is no "ping contributors"
 - this PR is sitting here without a reply for 2 months

leehambley:
@grosser those are features not bugs. And "sitting here without a reply for 2 months" is a side effect of this being a project people work on *in their free time*. 

Back off and stop being so abrasive in your online communications please, we owe you nothing, and you owe us nothing, save perhaps a little courtesy.

grosser:
After how many months is ok to ask for a review ?
You call me "incredibly impolite" for mentioning people by name, when I did that a thousand times already and have that done to me 10x more, and nobody before has complained about this, and I'm the one being "abrasive" ?

grosser:
doubt I can convince you, so I'm done here 👋

PR #463, opened 2019-07-01
Fix functional tests

mattbrictson:
I noticed that functional tests (`rake test:functional`) are failing on master. This PR fixes them with the following commits:

1. Strict host key checking causes functional tests to fail if the vagrant SSH server is not yet in the developer's known_hosts file. To avoid this problem, explicitly disable host key checking for functional tests.
2. PR #453 changed how commands were escaped but did not update the functional tests to match. Two functional tests were failing as a result. This commit fixes those tests.

I also made a change to `.travis.yml` as an optimization. Normally Travis CI runs builds twice on PRs, once for the push to the branch, and another for the PR itself. This is redundant and causes PRs checks to take longer that necessary. Fix this by restricting "push" builds to the master branch only.

No CHANGELOG entry since these are not user-facing changes.

### DIFF
--- a/test/functional/backends/test_netssh.rb
+++ b/test/functional/backends/test_netssh.rb
@@ -96,21 +96,18 @@ module SSHKit
       end
 
       def test_ssh_option_merge
-        verify_host_opt = if Net::SSH::Version::MAJOR >= 5
-                            { verify_host_key: :always }
-                          else
-                            { paranoid: true }
-                          end
-        a_host.ssh_options = verify_host_opt
+        keepalive_opt = { keepalive: true }
+        test_host = a_host.dup
+        test_host.ssh_options = keepalive_opt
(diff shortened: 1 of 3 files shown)


test/functional/backends/test_netssh.rb
```diff
@@ -96,21 +96,18 @@ def test_capture
...
 
       def test_ssh_option_merge
-        verify_host_opt = if Net::SSH::Version::MAJOR >= 5
-                            { verify_host_key: :always }
-                          else
-                            { paranoid: true }
-                          end
-        a_host.ssh_options = verify_host_opt
+        keepalive_opt = { keepalive: true }
+        test_host = a_host.dup
```
mattbrictson: I added `.dup` here because the modifications to `a_host` were polluting other tests.

Issue #464, opened 2019-07-01
Deploy failing after bundle update sshkit (1.19.0)


Fudoshiki:
```bash
$ cap production deploy
rvm 1.29.8-next (master) by Michal Papis, Piotr Kuczynski, Wayne E. Seguin [https://rvm.io]
ruby-2.6.3
ruby 2.6.3p62 (2019-04-16 revision 67580) +JIT [x86_64-linux]
00:00 git:wrapper
      01 mkdir -p /tmp
    ✔ 01 user@example.com 0.470s
      Uploading /tmp/git-ssh-app-production-user.sh 100.0%
      02 chmod 700 /tmp/git-ssh-app-production-user.sh
    ✔ 02 user@example.com 0.455s
00:02 git:check
      01 git ls-remote git@github.com:user/repo.git HEAD
      01 bcfd58198a1374c7592bf706f22698621c37af41 HEAD
    ✔ 01 user@example.com 1.061s
00:03 deploy:check:directories
      01 mkdir -p ~/repository/shared ~/repository/releases
    ✔ 01 user@example.com 0.495s
00:03 deploy:check:linked_dirs
      01 mkdir -p ~/repository/shared/log ~/repository/shared/tmp/pids ~/repository/shared/tmp/cache ~/repository/shared/tmp/sockets ~/repository/shared/public/assets
    ✔ 01 user@example.com 0.454s
00:04 deploy:check:make_linked_dirs
      01 mkdir -p ~/repository/shared/config/credentials
    ✔ 01 user@example.com 0.494s
#<Thread:0x0000000106d31dd0@/Users/user/.rvm/gems/ruby-2.6.3/gems/sshkit-1.19.0/lib/sshkit/runners/parallel.rb:10 run> terminated with exception (report_on_exception is true):
Traceback (most recent call last):
  10: from /Users/user/.rvm/gems/ruby-2.6.3/gems/sshkit-1.19.0/lib/sshkit/runners/parallel.rb:12:in `block (2 levels) in execute'
   9: from /Users/user/.rvm/gems/ruby-2.6.3/gems/sshkit-1.19.0/lib/sshkit/backends/abstract.rb:31:in `run'
   8: from /Users/user/.rvm/gems/ruby-2.6.3/gems/sshkit-1.19.0/lib/sshkit/backends/abstract.rb:31:in `instance_exec'
[...]

PR #465, opened 2019-07-01
allow tilde as home in paths

grosser:
fixes https://github.com/capistrano/sshkit/issues/464

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#465](https://github.com/capistrano/sshkit/pull/456): Fix withing/in escaping ~ - [@grosser](https://github.com/grosser)
 
 ## [1.19.0][] (2019-06-30)
 
--- a/lib/sshkit/backends/abstract.rb
+++ b/lib/sshkit/backends/abstract.rb
@@ -82,9 +82,10 @@ module SSHKit
 
       def within(directory, &_block)
         (@pwd ||= []).push directory.to_s
+        escaped = Command.shellescape_except_tilde(pwd_path)
         execute <<-EOTEST, verbosity: Logger::DEBUG
-          if test ! -d #{File.join(@pwd).shellescape}
-            then echo "Directory does not exist '#{File.join(@pwd).shellescape}'" 1>&2
+          if test ! -d #{escaped}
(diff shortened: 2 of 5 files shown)


test/unit/test_command.rb
```diff
@@ -84,6 +84,11 @@ def test_working_in_a_given_directory
       assert_equal "cd /opt/sites && /usr/bin/env ls -l", c.to_command
     end
 
+    def test_working_in_home_directory
```
mattbrictson: Nice tests!

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#465](https://github.com/capistrano/sshkit/pull/456): Fix withing/in escaping ~ - [@grosser](https://github.com/grosser)
```
mattbrictson: 🙌

lib/sshkit/backends/abstract.rb
```diff
@@ -82,9 +82,10 @@ def execute(*args)
 
       def within(directory, &_block)
         (@pwd ||= []).push directory.to_s
+        escaped = pwd_path.shellescape.gsub("\\~", "~")
```
mattbrictson: What do you think about extracting this double-escape logic into a separate utility method for clarity? The method could be called e.g. `shellescape_except_tilde` with a documentation comment explaining why it is desirable. Then we could reuse that method here and in `command.rb`. I am not sure where such a method would go since we don't really have a "utils" module; perhaps it could just be on the `SSHKit` module itself, for lack of a more obvious location.

lib/sshkit/backends/abstract.rb
```diff
@@ -82,9 +82,10 @@ def execute(*args)
...
         execute <<-EOTEST, verbosity: Logger::DEBUG
-          if test ! -d #{File.join(@pwd).shellescape}
-            then echo "Directory does not exist '#{File.join(@pwd).shellescape}'" 1>&2
+          if test ! -d #{escaped}
+            then echo "Directory does not exist #{escaped}" 1>&2
```
mattbrictson: The removal of the single quotes is a very minor change but it does break a functional test.

Before:

```
Directory does not exist '/tmp'
```

After:

```
Directory does not exist /tmp
```

Would you mind restoring the quotes?

grosser:
all done

PR #466, opened 2019-07-04
improve readme example

grosser:
- use more obvious `RAILS_ENV: :production` old one made me wonder "what does this even do, is this the ENV why is it not uppercase"
 - use more obvious `execute :rake`, misleading users into thinking all kinds of commands are prefdefined is bad, show them the general purpose tools
 - reduce example from 3->2 that's enough to show how it's done
 - remove `wait: 5` rarely used argument and users can just use builtin `sleep 5` if they want it
 - use simpler array syntax to not confuse

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,29 +6,31 @@ more servers.
 [![Gem Version](https://badge.fury.io/rb/sshkit.svg)](https://rubygems.org/gems/sshkit)
 [![Build Status](https://travis-ci.org/capistrano/sshkit.svg?branch=master)](https://travis-ci.org/capistrano/sshkit)
 
-## How might it work?
+## Example
 
-The typical use-case looks something like this:
+ - Connect to 2 servers
+ - Execute commands as `deploy` user with `RAILS_ENV=production`
+ - Execute commands in serial (default is `:parallel`)
 
 ```ruby
 require 'sshkit'
 require 'sshkit/dsl'
 include SSHKit::DSL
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -6,29 +6,30 @@ more servers.
...
-on %w{1.example.com 2.example.com}, in: :sequence, wait: 5 do |host|
+on ["1.example.com", "2.example.com"], in: :sequence do |host|
+  puts "Now executing on #{host}"
   within "/opt/sites/example.com" do
     as :deploy  do
-      with rails_env: :production do
-        rake   "assets:precompile"
-        runner "S3::Sync.notify"
-        execute :node, "socket_server.js"
+      with RAILS_ENV: :production do
```
mattbrictson: Non-blocking: To make it more clear that `:production` is not a special token, how about writing this instead as

```ruby
with RAILS_ENV: "production" do
```

?

README.md
```diff
@@ -6,29 +6,30 @@ more servers.
...
-## How might it work?
+## Example
 
-The typical use-case looks something like this:
+ - Connect to 2 servers
+ - Execute commands as `deploy` user with `RAILS_ENV=production`
+ - Execute commands in serial (default is `:parallel`)
 
 ```ruby
-require 'sshkit'
```
mattbrictson: This `require` is needed, otherwise the example fails with:

```
NameError (uninitialized constant SSHKit::DSL::Coordinator)
```

grosser:
thx, fixed

Issue #456, opened 2019-03-01
allow silencing upload/download


grosser:
would be great if I could pass `verbose: false` or so to these to make things a little less noisy ... atm I'm using a global silence which is not threadsafe ...

```ruby
  def silence_sshkit
    old = SSHKit.config.output_verbosity
    SSHKit.config.output_verbosity = :WARN
    yield
  ensure
    SSHKit.config.output_verbosity = old
  end
```

will-in-wi:
That would make sense to me. Would you be willing to attempt a patch?

grosser:
I'll take a look

PR #468, opened 2019-07-08
add a runnable example and document how verbosity/silence works

grosser:
fixes https://github.com/capistrano/sshkit/issues/456 since `verbosity: :DEBUG` does what I need
also fixed a few headings since ordering was level-2 -> level-4

@will-in-wi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#468](https://github.com/capistrano/sshkit/pull/468): Make `upload!` take a `:verbosity` option like `exec` does - [@grosser](https://github.com/grosser)
 
 ## [1.19.1][] (2019-07-02)
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ You can pass one or more hosts as parameters; this runs commands via SSH. Altern
 pass `:local` to run commands locally. By default SSKit will run the commands on all hosts in
 parallel.
 
-#### Running commands
+### Running commands
 
 All backends support the `execute(*args)`, `test(*args)` & `capture(*args)` methods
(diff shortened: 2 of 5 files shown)


grosser:
hmmm the "Finished in" line still appears I'll try to remove that too ...

grosser:
ah nvm ... but upload needs a tweak

grosser:
ok, upload uses the passed `verbosity:` now too ... looks good ?

examples/simple_connection.rb
```diff
@@ -0,0 +1,10 @@
+# tiny example so you can play with the sshkit or make a failing example for an issue
+require 'bundler/setup'
```
mattbrictson: Why is this needed?

grosser: it's meant as a starting point to debug sshkit itself, so it needs to use the local checkout and not the installed gem

CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ appear at the top.
 ## [Unreleased][]
 
   * Your contribution here!
+  * [#468](https://github.com/capistrano/sshkit/pull/468): Make `upload!` take a `:verbosity` option like `exec` does - [@grosser](https://github.com/grosser)
```
mattbrictson: 👍

README.md
```diff
@@ -115,6 +115,11 @@ the raised error.
 Helpers such as `runner()` and `rake()` which expand to `execute(:rails, "runner", ...)` and
 `execute(:rake, ...)` are convenience helpers for Ruby, and Rails based apps.
 
+### Verbosity / Silence
+
+ - raise verbosity of a command: `execute "echo DEAD", verbosity: :ERROR`
+ - hide a command from output: `execute "echo HIDDEN", verbosity: :DEBUG`
```
mattbrictson: Thanks for documenting this!

PR #469, opened 2019-07-20
allow home by env as $HOME in paths

Madogiwa0124:
Added support for $HOME in relation to #464, #465

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ appear at the top.
 
   * Your contribution here!
   * [#468](https://github.com/capistrano/sshkit/pull/468): Make `upload!` take a `:verbosity` option like `exec` does - [@grosser](https://github.com/grosser)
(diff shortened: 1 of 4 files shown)


leehambley:
Is there a very good reason why you cannot use `~`?

Expanding `~` is a Bash (and other shells) feature https://www.gnu.org/software/bash/manual/html_node/Tilde-Expansion.html which is distinct from the expansion of environment variables.

Whilst you may find it convenient to permit `$HOME` in paths, many people may be distressed and confused when finding out that it does not in fact do environment variable expansion and is simply a string replacement alias.

I prefer not to support confusing APIs, and would prefer not to merge this contribution, but I am open to discussion about the relative merits.

Thanks, and sorry it took me some time to get back to you, I've been travelling for work.

Madogiwa0124:
@leehambley 

I see 😄 

> Whilst you may find it convenient to permit $HOME in paths, many people may be distressed and confused when finding out that it does not in fact do environment variable expansion and is simply a string replacement alias.

Since I used $ HOME for my service, I created a PR, I agree with your opinion.
I fix my service to use a tilde. 👍 

Thanks! ✨

leehambley:
Thank you @Madogiwa0124 - I look forward to future contributions from you :) Thank you for using and trusting SSHKit :bowing_man:

Issue #470, opened 2019-07-30
Remove whoami check in as command


abhishiv:
Ref: #167 which is closed

PR #471, opened 2019-07-30
skip whoami check in as command

abhishiv:
ref #470

### DIFF
--- a/lib/sshkit/backends/abstract.rb
+++ b/lib/sshkit/backends/abstract.rb
@@ -112,13 +112,13 @@ module SSHKit
           @group = nil
         end
         if !@skip_whoami
-        execute <<-EOTEST, verbosity: Logger::DEBUG
-          if ! sudo -u #{@user.to_s.shellescape} whoami > /dev/null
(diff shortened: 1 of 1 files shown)


leehambley:
Can you clarify exactly why you want to introduce this please?

Another configuration option also needs docs and an explanation of why this exists and why we might need it. 

SSHKit has existed for six years without this bypass and only two people have ever raised issue with the way it is designed. (and I know for a fact that many people, my clients included) run Capistrano via SSHKit with restricted user accounts.

I'd love to understand more about your use-case before taking a general solution.

Thanks for taking the time to submit a PR.

abhishiv:
Hey @leehambley 

I think the issue is that many distributions(like ubuntu on digitalocean) come without sudo. Hence it requires first to install sudo using apt before you can use sshkit. 
 
Hence I think it's make more sense to remove this check. Is there a specific rationale that you want to keep this check for?

leehambley:
Applied "stale" label, will close in a few weeks if still stale.

PR #477, opened 2019-12-08
Removed dead links from README

ssimontis:
The links to rubyplus.com are no longer functioning. It appears that the domain expired and is no longer a Ruby site.


mattbrictson:
Good catch. Thanks!

PR #489, opened 2021-01-18
Add latest rubies to CI, including 3.0; drop rubinius CI support

mattbrictson:
This PR makes some minor modifications to our CI setup:

1. Add Ruby 2.6, 2.7, 3.0 to Travis CI.
2. Drop special gems from our Gemfile that supported Rubinius in CI. We haven't actually run Rubinius in CI since 2014, so these gems are definitely not needed now. One of the gems, `rubysl`, was preventing CI from working with Ruby 3.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: ruby
 rvm:
+  - 3.0
+  - 2.7
+  - 2.6
   - 2.5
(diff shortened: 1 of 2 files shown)


leehambley:
Nice one @mattbrictson , merged :muscle:

PR #509, opened 2023-02-06
Add functional test with vagrant on CI

colorbox:
This PR adds functional test on CI withs `vagrant up` on GitHub Actions. Some VM settings updated for keep alive connection for `test_upload_large_file`.

refs: https://github.com/jonashackt/vagrant-github-actions


mattbrictson (approved):
Awesome! 😻 

I am a little bit worried that the slow functional tests will burn through our GitHub Actions budget, but we can deal with that later, if/when it becomes a problem. For now, this is a much-appreciated improvement to our CI coverage.

PR #516, opened 2023-07-01
Fix CI failures by running Ruby <= 2.2 on older Ubuntu image

mattbrictson:
Old versions of Ruby (<= 2.2) no longer work on the latest Ubuntu image.

Fix by using `ubuntu-20.04` instead of `ubuntu-latest` for old Rubies.

Also create roll-ups job called `test-all` and `functional-all` that collects the results of all the test and functional job runs for various Ruby versions into two checks that we can use for the branch protection rule.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,7 +8,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ "2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6", "2.7", "3.0", "3.1", "3.2", "head" ]
+        ruby: ["2.3", "2.4", "2.5", "2.6", "2.7", "3.0", "3.1", "3.2", "head"]
     steps:
       - uses: actions/checkout@v3
       - name: Set up Ruby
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,14 +1,14 @@
...
-    branches: [ master ]
+    branches: [master]
   pull_request:
 jobs:
   test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: [ "2.0", "2.1", "2.2", "2.3", "2.4", "2.5", "2.6", "2.7", "3.0", "3.1", "3.2", "head" ]
+        ruby: ["2.3", "2.4", "2.5", "2.6", "2.7", "3.0", "3.1", "3.2", "head"]
```
mattbrictson, line 11: 🗒️ `"2.0", "2.1", "2.2"` have been moved to the new `test-legacy` job.

.github/workflows/ci.yml
```diff
@@ -19,6 +19,33 @@ jobs:
       - name: Run tests
         run: bundle exec rake test:units
 
+  test-legacy:
+    runs-on: ubuntu-20.04
```
mattbrictson, line 23: 🗒️ Use the older Ubuntu runner that is compatible with Ruby `"2.0", "2.1", "2.2"`

.github/workflows/ci.yml
```diff
@@ -19,6 +19,33 @@ jobs:
...
+      - uses: actions/checkout@v3
+      - name: Set up Ruby
+        uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: ${{ matrix.ruby }}
+          bundler-cache: true
+      - name: Run tests
+        run: bundle exec rake test:units
+
+  test-all:
```
mattbrictson, line 37: 🗒️ This aggregates all of the `test*` the matrix runs into a single check. That way the branch protection rule becomes much simpler.

.github/workflows/ci.yml
```diff
@@ -58,3 +99,15 @@ jobs:
 
       - name: Run functional tests
         run: bundle exec rake test:functional
+
+  functional-all:
```
mattbrictson, line 103: 🗒️ Likewise this aggregates all the `functional*` matrix runs into one check.

PR #524, opened 2023-12-22
Allow SFTP to be used for upload!/download! instead of SCP

### LINKED ISSUE #15: Allow to choose between sftp and scp.
Because well, it simply makes sense.

mattbrictson:
SSHKit can now use SFTP for file transfers. The default is still SCP, but this can be changed to SFTP per host:

```ruby
host = SSHKit::Host.new('user@example.com')
host.transfer_method = :sftp
```

or globally:

```ruby
SSHKit::Backend::Netssh.configure do |ssh|
  ssh.transfer_method = :sftp
end
```

Fixes #15

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -162,6 +162,23 @@ end
 In this case the `recursive: true` option mirrors the same options which are
 available to [`Net::{SCP,SFTP}`](http://net-ssh.github.io/net-scp/).
 
+## Set the upload/download method (SCP or SFTP).
+
+SSHKit can use SCP or SFTP for file transfers. The default is SCP, but this can be changed to SFTP per host:
+
+```ruby
+host = SSHKit::Host.new('user@example.com')
+host.transfer_method = :sftp
+```
+
+or globally:
+
+```ruby
+SSHKit::Backend::Netssh.configure do |ssh|
+  ssh.transfer_method = :sftp
+end
+```
+
 ## Setting global SSH options
 
 Setting global SSH options, these will be overwritten by options set on the
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -200,10 +200,17 @@ module SSHKit
       end
 
       def with_transfer(summarizer)
-        require_relative "netssh/scp_transfer"
+        transfer_method = host.transfer_method || self.class.config.transfer_method
+        transfer_class = if transfer_method == :sftp
+                           require_relative "netssh/sftp_transfer"
+                           SftpTransfer
+                         else
+                           require_relative "netssh/scp_transfer"
+                           ScpTransfer
+                         end
 
         with_ssh do |ssh|
-          yield(ScpTransfer.new(ssh, summarizer))
+          yield(transfer_class.new(ssh, summarizer))
         end
       end
     end
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -22,9 +22,26 @@ module SSHKit
   module Backend
 
     class Netssh < Abstract
+      def self.assert_valid_transfer_method!(method)
+        return if [nil, :scp, :sftp].include?(method)
+
+        raise ArgumentError, "#{method.inspect} is not a valid transfer method. Supported methods are :scp, :sftp."
+      end
+
       class Configuration
         attr_accessor :connection_timeout, :pty
+        attr_reader :transfer_method
         attr_writer :ssh_options
+
+        def initialize
+          self.transfer_method = :scp
+        end
+
+        def transfer_method=(method)
+          Netssh.assert_valid_transfer_method!(method)
+
+          @transfer_method = method
+        end
 
         def ssh_options
           default_options.merge(@ssh_options ||= {})
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -23,7 +23,7 @@ module SSHKit
 
     class Netssh < Abstract
       def self.assert_valid_transfer_method!(method)
-        return if [nil, :scp, :sftp].include?(method)
+        return if [:scp, :sftp].include?(method)
 
         raise ArgumentError, "#{method.inspect} is not a valid transfer method. Supported methods are :scp, :sftp."
       end
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -104,7 +104,7 @@ module SSHKit
         last_percentage = nil
         proc do |_ch, name, transferred, total|
           percentage = (transferred.to_f * 100 / total.to_f)
-          unless percentage.nan?
+          unless percentage.nan? || percentage.infinite?
             message = "#{action} #{name} #{percentage.round(2)}%"
             percentage_r = (percentage / log_percent).truncate * log_percent
             if percentage_r > 0 && (last_name != name || last_percentage != percentage_r)
--- a/lib/sshkit/backends/netssh.rb
+++ b/lib/sshkit/backends/netssh.rb
@@ -2,7 +2,6 @@ require 'English'
 require 'strscan'
 require 'mutex_m'
 require 'net/ssh'
-require 'net/scp'
 
 module Net
   module SSH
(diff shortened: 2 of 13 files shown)


mattbrictson:
@JasonPoll this PR is still a work in progress, but it should be ready for you to test. Let me know if this branch works with SFTP in your environment. Also would appreciate any suggestions/feedback on the implementation. Thanks!

will-in-wi (approved):
Code looks reasonable to me. Haven't tried it (I don't tend to use these methods…).

lib/sshkit/backends/netssh/sftp_transfer.rb
```diff
@@ -0,0 +1,46 @@
...
+          ssh.sftp.download!(remote, destination, options)
+          local ? true : destination.string
+        ensure
+          ssh.sftp.close_channel
+        end
+
+        def on_get(download, entry, offset, data)
+          entry.size ||= download.sftp.file.open(entry.remote, &:size)
+          summarizer.call(nil, entry.remote, offset + data.bytesize, entry.size)
+        end
```
mattbrictson, line 34: 🗒️ net-sftp calls `on_get` to provide download progress. This information is much different that how net-scp provides progress. In net-sftp, `offset` refers to the offset of the local destination, _before_ the bytes have been transferred. So some arithmetic is needed to calculate the progress: `offset + data.bytesize`.

Furthermore, the total size of the file being transferred, which is supposed to be provided via `entry.size`, is `nil`. To work around this, I had to explicitly make an sftp `open` call to read the file size.

lib/sshkit/backends/netssh/sftp_transfer.rb
```diff
@@ -0,0 +1,46 @@
...
+
+        def on_put(_upload, file, offset, data)
+          summarizer.call(nil, file.local, offset + data.bytesize, file.size)
+        end
```
mattbrictson, line 38: 🗒️ net-sftp calls `on_put` to provide upload progress. For uploads, `file.size` seems to be correct, so no workaround is needed.

test/functional/backends/netssh_transfer_tests.rb
```diff
@@ -0,0 +1,83 @@
+require 'securerandom'
+
+module SSHKit
+  module Backend
+    module NetsshTransferTests
```
mattbrictson, line 5: 🗒️ I extracted the upload/download tests into a module, so the same tests can easily be reused for both scp and sftp.

lib/sshkit/backends/netssh.rb
```diff
@@ -23,10 +22,27 @@ module SSHKit
   module Backend
 
     class Netssh < Abstract
+      def self.assert_valid_transfer_method!(method)
+        return if [nil, :scp, :sftp].include?(method)
```
JasonPoll, line 26: `nil`? 

I suppose it's inconsequential when `transfer_method` defaults to `:scp` along with the behaviors of `#with_transfer`.

mattbrictson: @JasonPoll `Host#transfer_method` is initially set to `nil` (which means "use the global default"), so I wanted to make sure that `nil` is an allowable value.

JasonPoll: ...oh wait...I think I understand. `nil` is considered a valid value because per-host configuration could be `nil` and global configuration could be what is specifying `:scp` or `:sftp`. 

Even in the case when both host and global configuration are assigned a `transfer_value` of `nil`, the `#with_transfer` method still embodies the previous behavior of transfers being performed via SCP.

mattbrictson: Maybe a more correct implementation would be to have `assert_valid_transfer_method!` strictly only allow :scp and :sftp, with a workaround in `Host#transfer_method=` to allow `nil`.

```ruby
# netssh.rb

def self.assert_valid_transfer_method!(method)
  return if [:scp, :sftp].include?(method)
  raise ArgumentError, "#{method.inspect} is not a valid transfer method. Supported methods are :scp, :sftp."
end
```

```ruby
# host.rb

def transfer_method=(method)
  Backend::Netssh.assert_valid_transfer_method!(method) unless method.nil?

  @transfer_method = method
end
```

mattbrictson: Addressed in 9bbcc2c

EXAMPLES.md
```diff
@@ -162,6 +162,23 @@ end
 In this case the `recursive: true` option mirrors the same options which are
 available to [`Net::{SCP,SFTP}`](http://net-ssh.github.io/net-scp/).
```
JasonPoll, line 163: confirmed this older documentation about `recursive: true` is still true with SFTP enabled.

lib/sshkit/backends/netssh/sftp_transfer.rb
```diff
@@ -0,0 +1,46 @@
...
+  module Backend
+    class Netssh < Abstract
+      class SftpTransfer
+        def initialize(ssh, summarizer)
+          @ssh = ssh
+          @summarizer = summarizer
+        end
+
+        def upload!(local, remote, options)
+          options = { progress: self }.merge(options || {})
```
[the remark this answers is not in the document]
JasonPoll: Thank you for the notes here -- when I had originally looked at this PR before heading out for the holidays I figured that I'd need to take a detour through Net::SFTP to figure out how the `summarizer` was getting the transfer progress information.

JasonPoll (approved):
I performed some functionality tests with a couple VMs, one with SCP enabled, and one without. My tests performed as expected. Detailed copy/pasted notes below.

> To keep my notes a bit more succinct, this is quick setup and testing method I used for testing `#upload!` and `#download!`: 
> 
> ```ruby
> require 'sshkit'
> require 'sshkit/dsl'
> include SSHKit::DSL
>    
> def doit(ip, &block)
>  puts "- global transfer method: :#{SSHKit::Backend::Netssh.config.transfer_method}"
>    on [ip] do |host|
>      yield host if block_given? 
>         
>      as :vagrant do 
>        within '/home/vagrant/tmp' do 
>          upload! 'junk.dat', 'junk.dat'
>          download! 'junk.dat', 'junk_downloaded.dat'
>        end 
>      end 
>    end 
>  end 
>  ```
> 
> To further make my notes succinct, redundant lines are eliminated/ellipsed. 


With everything set to SSHKit defaults, on `192.168.69.70` _(a host with SCP enabled)_ uploading and downloading via SCP works as expected: 

```ruby
doit('192.168.69.70')

- global transfer method: :scp
  INFO Uploading junk.dat 12.21%
  ...
  INFO Uploading junk.dat 100.0%
  INFO Downloading /home/vagrant/tmp/junk.dat 12.21%
  ...
  INFO Downloading /home/vagrant/tmp/junk.dat 100.0%
```

With everything set to defaults, on `192.168.69.42` _(a host with SCP disabled)_ it errors as expected when attempting to upload:

```ruby
doit('192.168.69.42')

- global transfer method: :scp
#<Thread:0x00007fc2d022f1d8 /home/vagrant/code/sshkit/lib/sshkit/runners/parallel.rb:10 run> terminated with exception (report_on_exception is true): 
...
SSHKit::Runner::ExecuteError: Exception while executing on host 192.168.69.42: SCP did not finish successfully (255): 
from /home/vagrant/code/sshkit/lib/sshkit/runners/parallel.rb:15:in `rescue in block (2 levels) in execute'
Caused by Net::SCP::Error: SCP did not finish successfully (255): 
from /home/vagrant/.rbenv/versions/3.2.2/lib/ruby/gems/3.2.0/gems/net-scp-4.0.0/lib/net/scp.rb:365:in `block (3 levels) in start_command'
```

Changing the global `transfer_method` to `:sftp` and the file uploads/downloads as expected: 

```ruby
SSHKit::Backend::Netssh.configure{|ssh| ssh.transfer_method = :sftp}
=> :sftp
doit('192.168.69.42')

- global transfer method: :sftp
  INFO Uploading junk.dat 10.94%
  ...
  INFO Uploading junk.dat 100.0%
  INFO Downloading junk_downloaded.dat 10.94%
  ...
  INFO Downloading junk_downloaded.dat 100.0%
```

Changing the `transfer_method` to `:sftp` on a per-host basis and the file uploads/downloads as expected: 

```ruby
doit('192.168.69.42') { |host| host.transfer_method = :sftp }

- global transfer method: :scp
  INFO Uploading junk.dat 10.94%
  ...
  INFO Uploading junk.dat 100.0%
  INFO Downloading junk_downloaded.dat 10.94%
  ...
  INFO Downloading junk_downloaded.dat 100.0%
```

This looks good to me!

lib/sshkit/backends/netssh.rb
```diff
@@ -64,16 +80,16 @@ def assign_defaults
       def upload!(local, remote, options = {})
         summarizer = transfer_summarizer('Uploading', options)
         remote = File.join(pwd_path, remote) unless remote.to_s.start_with?("/") || pwd_path.nil?
-        with_ssh do |ssh|
-          ssh.scp.upload!(local, remote, options, &summarizer)
+        with_transfer(summarizer) do |transfer|
+          transfer.upload!(local, remote, options)
```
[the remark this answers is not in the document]
JasonPoll: As someone who works in a 10+ year old code base as part of my day-job, I appreciate the approach you've taken here -- minimal changes to the long-established core of the codebase, refactoring previous functionality into a sort of configuration-based 'plugin' design, and adding new functionality with that same plugin functionality, all the while retaining backwards compatibility. 

💯, good sir.

mattbrictson:
Thanks @JasonPoll and @will-in-wi for the reviews! I plan on releasing this as part of sshkit 1.22.0 in mid-January, after 1.21.7 has some time in the wild. In the meantime, you can take advantage of the new sftp functionality by pointing your Gemfile at the master branch:

```ruby
gem "sshkit", github: "capistrano/sshkit", branch: "master"
```

PR #527, opened 2023-12-30
EXAMPLES.md corrections

JasonPoll:
While reviewing and testing @mattbrictson 's #524, I found myself referencing `EXAMPLES.md` and noted some inaccuracies. 

_Hoping it's ok to open this PR without a corresponding issue discussing this first._

### DIFF
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -154,7 +154,8 @@ end
 ```
 
 In this case the `recursive: true` option mirrors the same options which are
-available to [`Net::{SCP,SFTP}`](http://net-ssh.github.io/net-scp/).
+available to [`Net::SCP`](http://net-ssh.github.io/net-scp/) and
+[`Net::SFTP`](https://github.com/net-ssh/net-sftp).
 
 ## Setting global SSH options
 
--- a/EXAMPLES.md
+++ b/EXAMPLES.md
@@ -230,16 +230,16 @@ end
 
 ```ruby
 # The default format is pretty, which outputs colored text
-SSHKit.config.format = :pretty
+SSHKit.config.use_format :pretty
 
 # Text with no coloring
-SSHKit.config.format = :simpletext
+SSHKit.config.use_format :simpletext
 
 # Red / Green dots for each completed step
-SSHKit.config.format = :dot
+SSHKit.config.use_format :dot
 
 # No output
-SSHKit.config.format = :blackhole
+SSHKit.config.use_format :blackhole
 ```
 
 ## Implement a dirt-simple formatter class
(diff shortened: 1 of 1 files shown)


EXAMPLES.md
```diff
@@ -148,9 +145,6 @@ end
 This spares one from having to figure out the correct escaping sequences for
 something like "echo(:cat, '...?...', '> /etc/sudoers.d/yolo')".
 
-**Note:** The `upload!()` method doesn't honor the values of `within()`, `as()`
-etc, this will be improved as the library matures, but we're not there yet.
```
JasonPoll, line 152: This (and the above) appeared to be inaccurate based on [my testing of #524](https://github.com/capistrano/sshkit/pull/524#pullrequestreview-1799265301).

EXAMPLES.md
```diff
@@ -235,16 +230,16 @@ end
 
 ```ruby
 # The default format is pretty, which outputs colored text
-SSHKit.config.format = :pretty
+SSHKit.config.use_format :pretty
```
JasonPoll, line 233: These appeared to have never been updated in 2cb6326.

EXAMPLES.md
```diff
@@ -254,7 +249,7 @@ module SSHKit
   module Formatter
     class MyFormatter < SSHKit::Formatter::Abstract
       def write(obj)
-        case obj.is_a? SSHKit::Command
+        if obj.is_a? SSHKit::Command
```
JasonPoll, line 252: At least with Ruby 3.2.2, this `case` produced an error:

```
SyntaxError: unexpected `end', expecting `when'
```

EXAMPLES.md
```diff
@@ -160,7 +154,8 @@ end
 ```
 
 In this case the `recursive: true` option mirrors the same options which are
-available to [`Net::{SCP,SFTP}`](http://net-ssh.github.io/net-scp/).
+available to [`Net::SCP`](http://net-ssh.github.io/net-scp/) and
```
mattbrictson, line 157: @JasonPoll can you change the SCP link to https://github.com/net-ssh/net-scp ? It looks like http://net-ssh.github.io/net-scp/  is out of date (the header says "Last Update: 2014").

EXAMPLES.md
mattbrictson: Good catch!

mattbrictson (changes requested):
Thanks for this! One small change requested; see comments

EXAMPLES.md
JasonPoll: Can do.

mattbrictson (approved):

PR #534, opened 2024-04-04
Avoid calling closed? outside of synchronize block

djmb:
`closed?` calls `process` on the connection which is not safe because we have not synchronised the connection pool. Another thread might concurrently checkout the connection and start sending commands as well.

### DIFF
--- a/lib/sshkit/backends/connection_pool/cache.rb
+++ b/lib/sshkit/backends/connection_pool/cache.rb
@@ -36,7 +36,7 @@ class SSHKit::Backend::ConnectionPool::Cache
   def evict
     # Peek at the first connection to see if it is still fresh. If so, we can
     # return right away without needing to use `synchronize`.
-    first_expires_at, first_conn = connections.first
+    first_expires_at, _first_conn = connections.first
     return if (first_expires_at.nil? || fresh?(first_expires_at))
 
     connections.synchronize do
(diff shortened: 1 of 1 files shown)


djmb:
When using SSHKit with a 100+ hosts, we sometimes get IOErrors like this:

```
[ERROR (IOError): Exception while executing on host foo: closed stream]
```

Debugging showed that a packet was being sent to the remove server twice, which was causing it to close the connection. This was caused by two threads using the connection concurrently, the eviction thread and on of the SSHKit parallel runner threads. 

The duplicate packets came from [here](https://github.com/net-ssh/net-ssh/blob/eb2a49db51e33829e102e8c651e8ae0edc3ad3b7/lib/net/ssh/buffered_io.rb#L99-L108) in net-ssh, where both threads call `send` before either calls `output.consume!`.

Looking at the eviction code, the call to `closed?(first_conn)` outside the synchronize block looked suspicious as it calls `conn.process(0)` which will send a test packet.

The check looks like a performance optimisation, so I think it should be safe to remove. We've not seen the errors again with this fix running.

mattbrictson (changes requested):
Thanks for the fix! Could you make the following change to address the rubocop failure in CI?

djmb:
Thanks for the review @mattbrictson - I've committed your suggestion.

mattbrictson (approved):

PR #536, opened 2024-05-16
Drop `mutex_m` dependency

Earlopain:
Just use a plain mutex instead since only `synchronize` is used.


mattbrictson (approved):
Excellent! Thanks for the PR 🙏

Issue #537, opened 2024-05-19
Unbracketed IPv6 addresses don't work


jeromedalbert:
### Problem

I can use SSHKit with a bracketed IPv6 address like so:

```ruby
on ['[1fff:0:a88:85a3::ac1f]'] do |host|
```

But if I try to use an unbracketed IPv6 address

```ruby
on ['1fff:0:a88:85a3::ac1f'] do |host|
```

I get the following error:

<code>/Users/jerome/.asdf/installs/ruby/3.3.1/lib/ruby/gems/3.3.0/gems/sshkit-1.22.2/lib/sshkit/runners/parallel.rb:15:in `rescue in block (2 levels) in execute': Exception while executing on host 1fff: getaddrinfo: nodename nor servname provided, or not known (SSHKit::Runner::ExecuteError)</code>

### Additional information

Ubracketed IPv6 addresses are valid IPv6 addresses, so I think SSHKit should be able to parse them.

In my particular case, I am encountering this issue when using [Kamal](https://github.com/basecamp/kamal). Supplying an unbracketed IPv6 address in the Kamal `config/deploy.yml` configuration doesn't work because Kamal then passes that IP to SSHKit, as explained in this open issue: https://github.com/basecamp/kamal/issues/588. As mentioned in that issue, using brackets as a workaround doesn't work out of the box and requires further hacks overriding SSHKit's behavior, but these hacks all stem from SSHKit not properly parsing unbracketed IPv6 in the first place. Parsing unbracketed IPv6 addresses correctly in SSHKit would naturally solve those issues and allow Kamal to support IPv6 addresses, unbracketed and probably bracketed too.

### Related issues

- https://github.com/basecamp/kamal/issues/588, as mentioned above
- https://github.com/capistrano/sshkit/issues/403, fixed by https://github.com/capistrano/sshkit/pull/460, but still requires brackets

mattbrictson:
Closed via #538

PR #538, opened 2024-05-19
Handle unbracketed ipv6 addresses

jeromedalbert:
Potential fix for https://github.com/capistrano/sshkit/issues/537.

### Solution

- Add a new `IPv6HostParser`. This parser has a high priority in the parser list so IPv6 addresses are not incorrectly picked up and parsed by `HostWithPortParser`'s  simple regex.
- Because this parser has a high priority, we need its regex to be as strict and narrow as possible to prevent it from picking up and incorrectly parsing non-IPv6 addresses. That's why it uses `Resolv::IPv6::Regex` from the stdlib.

### Notes

My [previous solution](https://github.com/capistrano/sshkit/pull/538/commits/246630dec38013a282e50aafe0d7d747385bd301) tweaked the regex of the existing `IPv6HostWithPortParser`, but this resulted in a regression for local hostnames, so I reverted it and used `Resolv::IPv6::Regex` instead, which should prove more stable. I added an extra unit test to validate the local hostname scenario.


mattbrictson (changes requested):
Thanks for the PR! Sorry for the slow response.

I think the ambiguity in the IPv6 regex means that moving it up in the parser priority will introduce problems. For example, there a regression when the hostname is "db" and a port is specified:

```ruby
# Expected behavior
Host.new("db:22")
# => hostname="db", port=22

# Regression
Host.new("db:22")
# => hostname="db:22", port=nil
```

Is there a way to make the regex more strict to solve this? I am not very familiar with IPv6 addresses, so I'm not sure I can offer a solution, other than to stick with the status-quo `[]` requirement.

Maybe using `Resolv::IPv6::Regex` (as you originally implemented) is the way to go?

jeromedalbert:
@mattbrictson Good point. Because the IPv6 parser still needs a high priority, we need a strict IPv6 regex. `Resolv::IPv6::Regex` seems to be the way to go, it is pretty [involved](https://github.com/ruby/ruby/blob/73b7eebf07a5045a47bff72a7fc70e5e6aee7b94/lib/resolv.rb#L2966-L3030) in order to catch all the different IPv6 formats, so that should be plenty strict for our needs.

I have updated the PR and description accordingly and added an extra unit test for the `db:22` scenario you mentioned.

test/unit/test_host.rb
```diff
@@ -28,6 +28,12 @@ def test_host_with_port
       assert_equal 'example.com', h.hostname
     end
 
+    def test_custom_host_with_port
+      h = Host.new 'db:22'
```
mattbrictson, line 32: ✨ Thanks for the additional test!

mattbrictson (approved):
Looks great!

PR #522, opened 2023-11-17
Re-enable functional tests in CI

mattbrictson:
Functional tests were disabled in CI in #523 due to GitHub Actions issues.

This PR plans to reenable functional tests in CI once those issues are resolved. It also cargo-cults some Vagrantfile improvements from https://github.com/fastapi-mvc/example/blob/6b29b3fe3dc9004aa7492c4d2ecd569fdf7e3ec3/Vagrantfile

PR #539, opened 2024-06-15
Replace Vagrant with Docker Compose for running functional tests

mattbrictson:
Our functional tests had been failing in CI for quite some time. They relied on VirtualBox (managed by Vagrant), which no longer works reliably on the standard GitHub Actions runners. We disabled functional tests in CI due to this problem (see #523).

This PR fixes the functional tests by swapping out Vagrant for Docker Compose, following the example of https://github.com/capistrano/capistrano/pull/2159.

Docker Compose works fine with GitHub Actions, so I also reenabled functional tests in CI, borrowing from #522.

And they work now!

Screenshot 2024-06-14 at 5 24 06 PM


.github/workflows/ci.yml
```diff
@@ -69,3 +69,30 @@ jobs:
           bundler-cache: true
       - name: Run rubocop
         run: bundle exec rake lint
+
+  functional:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ruby: ["2.0", "ruby"]
```
mattbrictson, line 77: 🗒️ `"ruby"` in this context means the latest stable version of Ruby. As of this writing, it is 3.3.3. Using this keyword means the functional tests will always run against the latest version of Ruby, without us having to manually update the workflow.

Rakefile
```diff
@@ -21,10 +21,6 @@ namespace :test do
 
 end
 
-Rake::Task["test:functional"].enhance do
-  warn "Remember there are still some VMs running, kill them with `vagrant halt` if you are finished using them."
```
mattbrictson, line 25: 🗒️ `DockerWrapper` will automatically start Docker Compose on demand and clean it up after tests are done, so this instruction is no longer relevant.

test/boxes.json
```diff
@@ -1,17 +0,0 @@
-[
```
mattbrictson, line 1: 🗒️ This JSON file defined 3 different host VMs. However, in practice, only 1 of these was ever used in the functional tests. This was extra complexity wasn't serving a purpose, so I removed the JSON in favor of just hardcoding a single host.

test/functional/test_ssh_server_comes_up_for_functional_tests.rb
```diff
@@ -1,24 +0,0 @@
...
-
-    def host
-      @_host ||= Host.new('')
-    end
-
-    def test_that_it_works
-      assert true
-    end
-
-    def test_creating_a_user_gives_us_back_his_private_key_as_a_string
```
mattbrictson, line 15: 🗒️ This seemed to be testing the Vagrant infrastructure itself, as opposed to testing functionality of SSHKit. It doesn't serve a purpose with Docker, so I removed it.

test/support/docker_wrapper.rb
```diff
@@ -0,0 +1,50 @@
+Minitest.after_run do
+  DockerWrapper.stop if DockerWrapper.running?
+end
```
mattbrictson, line 3: 🗒️ This ensures that the container is cleaned up at exit.

test/support/docker_wrapper.rb
```diff
@@ -0,0 +1,50 @@
...
+
+module DockerWrapper
```
mattbrictson, line 5: 🗒️ This code was mostly lifted from https://github.com/capistrano/capistrano/pull/2159.

PR #542, opened 2024-09-05
Add ostruct dependency to fix warning on Ruby 3.3.5+

### LINKED ISSUE #541: warning: ostruct was loaded from the standard library
Using Ruby 3.3.5 and sshkit 1.23.0, I get this warning:

```
/home/jason/.rbenv/versions/3.3.5/lib/ruby/gems/3.3.0/gems/sshkit-1.23.0/lib/sshkit.rb:29: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.
```

mattbrictson:
Fixes the following warning on Ruby 3.3.5+ and Ruby 3.4:

> sshkit-1.23.0/lib/sshkit.rb:29: warning: ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.
You can add ostruct to your Gemfile or gemspec to silence this warning.

Closes #541

### DIFF
--- a/sshkit.gemspec
+++ b/sshkit.gemspec
@@ -24,7 +24,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('net-ssh',  '>= 2.8.0')
   gem.add_runtime_dependency('net-scp',  '>= 1.1.2')
   gem.add_runtime_dependency('net-sftp', '>= 2.1.2')
-  gem.add_runtime_dependency('ostruct')
+  gem.add_runtime_dependency('ostruct') if RUBY_VERSION >= "2.5"
 
   gem.add_development_dependency('danger')
   gem.add_development_dependency('minitest', '>= 5.0.0')
--- a/sshkit.gemspec
+++ b/sshkit.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |gem|
   gem.add_runtime_dependency('net-ssh',  '>= 2.8.0')
   gem.add_runtime_dependency('net-scp',  '>= 1.1.2')
   gem.add_runtime_dependency('net-sftp', '>= 2.1.2')
+  gem.add_runtime_dependency('ostruct')
 
   gem.add_development_dependency('danger')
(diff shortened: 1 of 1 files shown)


tisba:
I also get (sshkit 1.23.0):

> /.gem/ruby/3.3.5/gems/sshkit-1.23.0/lib/sshkit.rb:29: warning: logger was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0.

There was an overview what is planned to change in terms of stdlib/bundled gems in upcoming Ruby versions, but I cannot find it :(

mattbrictson:
@tisba what version of net-ssh are you using? The only place I can find where sshkit is doing `require 'logger'` is in this conditional branch, which should only run for very old versions of net-ssh (< 4.0.0):

https://github.com/capistrano/sshkit/blob/892d5bf820c58c046d304b05ce1f7f13de99cda6/lib/sshkit/backends/netssh.rb#L68-L73

tisba:
net-ssh (7.2.3)

But sorry for the noise, I'm pretty sure it's net-ssh's fault. They don't have the `logger` gem as a dependency.

mattbrictson:
> There was an overview what is planned to change in terms of stdlib/bundled gems in upcoming Ruby versions, but I cannot find it

As of Ruby 3.3.5, here is the list of gems that will no longer be built into future Ruby versions: https://github.com/ruby/ruby/blob/ef084cc8f4958c1b6e4ead99136631bef6d8ddba/lib/bundled_gems.rb#L5-L35

E.g.

```ruby
"ostruct" => "3.5.0",
```

means that `ostruct` will be removed in Ruby 3.5.0, which is what produces this warning:

> ostruct was loaded from the standard library, but will no longer be part of the default gems starting from Ruby 3.5.0

dorianmariecom:
Can a new version be release? :) plz thx

tisba:
Already has been released as 1.23.1, see https://rubygems.org/gems/sshkit/versions/1.23.1

PR #545, opened 2024-10-10
Update to be frozen string literal safe

npezza93:


lib/sshkit/deprecation_logger.rb
```diff
@@ -11,8 +11,14 @@ def log(message)
       return if @out.nil?
       warning_msg = "[Deprecated] #{message}\n"
       caller_line = caller.find { |line| !line.include?('lib/sshkit') }
-      warning_msg << "    (Called from #{caller_line})\n" unless caller_line.nil?
-      @out << warning_msg unless @previous_warnings.include?(warning_msg)
+      warning_msg = "#{warning_msg}    (Called from #{caller_line})\n" unless caller_line.nil?
```
mattbrictson, line 14: I think this change isn't needed because `warning_msg` will not be frozen. This is because `warning_msg` is constructed as in interpolated string (i.e. "...#{...}..."), which are not frozen. I confirmed this in Ruby 3.4.0-preview2:

```
(rbenv local 3.4.0-preview2 && RUBYOPT=--enable=frozen-string-literal irb)
irb(main):001> "hi".frozen?
=> true
irb(main):002> "hi #{'there'}".frozen?
=> false
```

lib/sshkit/deprecation_logger.rb
```diff
@@ -11,8 +11,14 @@ def log(message)
...
+      unless @previous_warnings.include?(warning_msg)
+        if @out.frozen?
```
mattbrictson, line 16: I wonder if it would be better to simplify this logic and remove the `@out.frozen?` check? My understanding of `DeprecationLogger` is that it should be initialized with an output stream of some sort. By default, this is `$stderr`. It doesn't really make sense for the logger to be initialized with a frozen `@out` object, and I think it would be reasonable to raise a `FrozenError` in this case.

In other words I think all the changes in this file can safely be reverted.

lib/sshkit/color.rb
```diff
@@ -44,8 +44,8 @@ def colorize(obj, color, mode=nil)
       return string unless COLOR_CODES.key?(color)
 
       result = mode == :bold ? "\e[1;" : "\e[0;"
-      result << COLOR_CODES.fetch(color).to_s
-      result << ";49m#{string}\e[0m"
+
+      "#{result}#{COLOR_CODES.fetch(color)};49m#{string}\e[0m"
```
mattbrictson, line 48: 👍

.github/workflows/ci.yml
```diff
@@ -30,6 +30,8 @@ jobs:
           bundler-cache: true
       - name: Run tests
         run: bundle exec rake test:units
+        env:
+          RUBYOPT: ${{ startsWith(matrix.ruby, 'head') && '--enable=frozen-string-literal' || '' }}
```
mattbrictson, line 34: Nice enhancement!

test/unit/backends/test_abstract.rb
```diff
@@ -112,14 +112,14 @@ def test_within_home
       end
 
       def test_background_logs_deprecation_warnings
-        deprecation_out = ''
+        deprecation_out = []
```
mattbrictson, line 115: Changing this to an array slightly alters the implied API contract of the deprecation logger, which expects an output stream. A more lightweight change could be:

```suggestion
        deprecation_out = +''
```

This would create a mutable string instead of a frozen one.

Alternatively, to be more explicit about passing a stream-like object, we could do this:

```suggestion
        deprecation_out = StringIO.new
```

I think both would be preferable to passing an array.

mattbrictson (changes requested):
Thanks for the PR! I think it is important to make sshkit frozen-string-literal safe to support future versions of Ruby.

I suggest simplifying some of the changes somewhat. Please see my comments below.

test/unit/backends/test_abstract.rb
npezza93: Hmmm looks like +'' isnt a thing in ruby 2.0 and StringIO doesnt support << or lines. How would you like to proceed here? 

image

npezza93: Actually nevermind, think i figured it.

PR #549, opened 2025-01-12
Drop support for Ruby 2.0, 2.1, 2.2, 2.3, and 2.4

mattbrictson:


.github/workflows/ci.yml
```diff
@@ -77,7 +49,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        ruby: ["2.0", "ruby"]
+        ruby: ["2.5", "ruby"]
```
mattbrictson, line 52: 🗒️ Run slow functional tests on the oldest Ruby we support (2.5) and the latest Ruby.

.rubocop.yml
```diff
@@ -18,12 +18,12 @@ Style/ClassAndModuleChildren:
   Enabled: false
 Style/DoubleNegation:
   Enabled: false
-Style/FileName:
+Naming/FileName:
```
mattbrictson, line 21: 🗒️ I had to upgrade the rubocop gem to support `TargetRubyVersion: 2.5`. The new version migrated the names of some cops, hence the changes in these configs.

Gemfile
```diff
@@ -1,8 +1,3 @@
 source 'https://rubygems.org'
 
 gemspec
-
-# public_suffix 3+ requires ruby 2.1+
-if Gem::Requirement.new('< 2.1').satisfied_by?(Gem::Version.new(RUBY_VERSION))
```
mattbrictson, line 6: 🗒️ No longer needed, since we no longer support Ruby < 2.1.

lib/sshkit/backends/connection_pool.rb
```diff
@@ -1,5 +1,4 @@
 require "monitor"
-require "thread"
```
mattbrictson, line 2: 🗒️ RuboCop flagged this as an unnecessary `require` (the `Thread` constant is always available without needing to require it.)

test/functional/backends/test_netssh.rb
```diff
@@ -131,7 +131,7 @@ def test_execute_raises_on_non_zero_exit_status_and_captures_stdout_and_stderr
 
       def test_test_does_not_raise_on_non_zero_exit_status
         Netssh.new(a_host) do |_host|
-          test :false
+          test :false # rubocop:disable Lint/BooleanSymbol
```
mattbrictson, line 134: 🗒️ RuboCop thinks this is a typo, and that we mean `false` and not `:false`. However in this case `:false` is the name of a command we are executing via SSH, not a boolean value.

will-in-wi (approved):

